PR #2, opened 2016-11-21
Implementing Cormode-Muthukrishnan algorithm and t-digest algorithm for computing biased quantiles

xichen2020:
cc @robskillington @martin-mao @cw9 

The Cormode-Muthukrishnan algorithm is what statsite currently uses for estimating quantiles for timers and is generally more accurate when the number of samples is small. However the algorithm (and the underlying data structure in the implementation) does not allow merging, which means it can only consume raw data points and uses more network bandwidth.

The t-digest algorithm is slightly less accurate when the number of samples is small (e.g., < 100). Additionally it doesn't guarantee the estimate is always one of the values in the provided data stream since it's based on clustering and interpolation. However it does show similar accuracy when the number of samples is large. More importantly, it supports merging naturally.

This PR implements the Cormode-Muthukrishnan algorithm and t-digest algorithm for computing biased quantiles. Depending on the accuracy and bandwidth requirements, we can adopt one or the other, or a hybrid of both.


pool/bucketized.go
```diff
@@ -0,0 +1,112 @@
...
+	p.alloc = alloc
+}
+
+func (p *bucketizedObjectPool) Get(capacity int) interface{} {
+	if capacity > p.maxBucketCapacity {
+		return p.alloc(capacity)
+	}
+	for i := range p.buckets {
+		if p.buckets[i].capacity >= capacity {
+			return p.buckets[i].pool.Get().([]float64)
```
robskillington: Should set to zero size before returning, or on putting.  

Probably best on getting though.

pool/object.go
```diff
@@ -0,0 +1,110 @@
...
+
+type objectPoolMetrics struct {
+	free       tally.Gauge
+	total      tally.Gauge
+	getOnEmpty tally.Counter
+	putOnFull  tally.Counter
+}
+
+// NewObjectPool creates a new pool
+func NewObjectPool(opts ObjectPoolOptions) ObjectPool {
```
robskillington: We should probably move this to m3x as well from m3db.  I know the native pool will be ideal, but in case you do need to use a normal pool too, would be best to use a common one in m3x than a copy/paste into m3aggregator.

quantile/cm/heap.go
```diff
@@ -0,0 +1,79 @@
...
+func (h *minHeap) Pop() float64 {
+	var (
+		old = *h
+		n   = old.Len()
+		val = old[0]
+	)
+
+	old[0], old[n-1] = old[n-1], old[0]
+	h.heapify(0, n-1)
+	*h = (*h)[0 : n-1]
```
robskillington: Won't this allocate a new array here?  If you take the value of `[0:n-1]` it will copy into a newly allocated slice no?

quantile/cm/heap.go
```diff
@@ -0,0 +1,79 @@
...
+		parent := (i - 1) / 2
+		if parent == i || h[parent] <= h[i] {
+			break
+		}
+		h[parent], h[i] = h[i], h[parent]
+		i = parent
+	}
+}
+
+func (h minHeap) heapify(i, n int) {
```
robskillington: Any reason not to use the stdlib heap and just supply the correct `Less` logic to make it a min heap? https://golang.org/pkg/container/heap/

glide.yaml
```diff
@@ -0,0 +1,11 @@
+package: github.com/m3db/m3aggregator
+import:
+- package: github.com/m3db/m3db
+  subpackages:
+  - pool
+- package: github.com/uber-go/tally
```
xichen2020: sure will do

quantile/cm/list.go
```diff
@@ -0,0 +1,90 @@
...
+// THE SOFTWARE.
+
+package cm
+
+var (
+	emptySampleList sampleList
+)
+
+// sampleList is a list of samples
+type sampleList struct {
```
robskillington: What's the plan to be able to use native pool/etc for this?

pool/bucketized.go
xichen2020: Yeah this pool is meant to be used as the underlying object pool for `FloatsPool` in `quantile/cm/floats_pool.go` and `CentroidsPool` in `quantile/tdigest/centroids_pool.go` where we set to zero size on put.

pool/object.go
xichen2020: Agreed, @kobolog will move it over to m3x once the GC diff lands

quantile/cm/heap.go
xichen2020: No it shouldn't, it's just a slice copy

quantile/cm/heap.go
xichen2020: So the reason why this is done instead of using the standard heap is that every thing in the std heap is an interface{}, which means every time I push a float to the heap or pop a float off of the heap requires a conversion between interface{} and float, which means memory allocations so more GC overhead. By making the heap typed it should remove that memory overhead.

I'll add the reasoning as part of the comments above the minHeap type def.

quantile/cm/stream.go
```diff
@@ -0,0 +1,278 @@
...
+func (s *stream) addToBuffer(value float64) {
+	if s.numValues > 0 && value < s.insertPointValue() {
+		s.addToMinHeap(&s.bufLess, value)
+	} else {
+		s.addToMinHeap(&s.bufMore, value)
+	}
+}
+
+// insert inserts a sample into the stream
+func (s *stream) insert() {
```
robskillington: Before I read this I'm going to have to read the statsite implementation.  Might take a little time.

quantile/tdigest/options.go
```diff
@@ -0,0 +1,71 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package tdigest
+
+const (
+	defaultCompression = 100.0
```
robskillington: Any reason to use 0.0 to 100.0 instead of 0.0 to 1.0?  

Also, do we need a `Validate() error` method on this options too?

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+
+func unmergedCapacity(compression float64) int {
+	compression = math.Min(math.Max(20, compression), 1000)
+	return (int)(7.5 + 0.37*compression - 2e-4*compression*compression)
+}
+
+// NewTDigest creates a new t-digest
+// TODO(xichen): add pooling for t-digests
+// TODO(xichen): add metrics
+func NewTDigest(opts Options) TDigest {
```
robskillington: Going to read the reference t-digest implementation before reviewing this.

quantile/cm/heap.go
robskillington: Makes sense 👍

quantile/cm/sample.go
```diff
@@ -0,0 +1,40 @@
...
+	return &Sample{}
+}
+
+// reset resets a sample
+func (s *Sample) reset() {
+	*s = emptySample
+}
+
+// setData sets sample data
+func (s *Sample) setData(value float64, numRanks int64, delta int64) {
```
xichen2020: sure thing

quantile/cm/sample_pool.go
```diff
@@ -0,0 +1,46 @@
...
+		return newSample()
+	})
+}
+
+func (p *samplePool) Get() *Sample {
+	return p.pool.Get().(*Sample)
+}
+
+func (p *samplePool) Put(sample *Sample) {
+	p.pool.Put(sample)
```
xichen2020: yeah can call reset in get

quantile/cm/stream.go
```diff
@@ -0,0 +1,278 @@
...
+	compressCursor  *Sample    // compression cursor
+	compressMinRank int64      // compression min rank
+}
+
+// NewStream creates a new sample stream
+// TODO(xichen): the stream object itself should be pooled too
+// TODO(xichen): add metrics
+func NewStream(capacity int, opts Options) Stream {
+	samplePool := opts.SamplePool()
+	floatsPool := opts.FloatsPool()
```
xichen2020: I called it `floatsPool` to follow the naming convention for slice pools (e.g.,`BytesPool`). Can call it `floatSlicePool` to make it more explicit

quantile/cm/stream.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return &stream{
+		capacity:   capacity,
+		eps:        opts.Eps(),
+		quantiles:  opts.Quantiles(),
+		bufLess:    minHeap(floatsPool.Get(capacity)),
+		bufMore:    minHeap(floatsPool.Get(capacity)),
+		samplePool: samplePool,
+		floatsPool: floatsPool,
+	}
```
xichen2020: sounds good

quantile/cm/stream.go
xichen2020: sure thing

quantile/tdigest/options.go
xichen2020: `Compression` is not necessarily smaller than 1.0 according to the paper.

Can add `Validate()`

quantile/cm/stream.go
robskillington: Ah right cause its an array/slice.  Makes sense, I retract my comment.

quantile/cm/heap_test.go
```diff
@@ -0,0 +1,40 @@
...
+	iter := 10
+	for i := iter - 1; i >= 0; i-- {
+		h.Push(float64(i))
+		require.Equal(t, iter-i, h.Len())
+	}
+	for i := 0; i < iter; i++ {
+		require.Equal(t, float64(i), h.Min())
+		require.Equal(t, float64(i), h.Pop())
+	}
+}
```
robskillington: Might be worth adding a few more min heap tests?  We could potentially take the cases already written for the golang stdlib heap and just simply write up equivalent cases.  It should be pretty easy to translate them I should think?  Don't need all of them but having good coverage would be ideal.

robskillington:
Just reviewed Cormode-Muthukrishnan implementation, looks good 👍 

Next reviewing tdigest implementation.

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// mergedCapacity computes the capacity of the merged centroid slice
+func mergedCapacity(compression float64) int {
+	return int(math.Pi*compression/2 + 0.5)
+}
+
+func unmergedCapacity(compression float64) int {
+	compression = math.Min(math.Max(20, compression), 1000)
+	return (int)(7.5 + 0.37*compression - 2e-4*compression*compression)
```
robskillington: Do you want to add a comment regarding this formula was taken from an existing implementation and that it was created by regressing against known sizes for sample compression values?

robskillington: Also: any reason to cast instead of `int(...)`?

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+	mergedWeight   float64    // total weight of merged centroids
+	unmerged       []Centroid // unmerged centroid slice capacity
+	unmergedWeight float64    // total weight of unmerged centroids
+	minValue       float64    // minimum value
+	maxValue       float64    // maximum value
+}
+
+// mergedCapacity computes the capacity of the merged centroid slice
+func mergedCapacity(compression float64) int {
+	return int(math.Pi*compression/2 + 0.5)
```
xichen2020: The comment in the `MergingDigest` implementation says `should only need ceiling(compression * PI / 2)` so I'd rather start with that :) worst case we'll reallocate.

`math.Ceil(X)` will always round X up, whereas `int(X + 0.5)` only rounds up if the fractional part of X is bigger than 0.5

quantile/tdigest/tdigest.go
xichen2020: yeah can add a comment, not sure I understand the comment about casting?

quantile/tdigest/tdigest.go
robskillington: Does it matter here?  Shouldn't we err on the side of allocating just a tiny bit more so using `math.Ceil(...)` seems more ideal here?

quantile/tdigest/tdigest.go
robskillington: i.e. any reason not to use the more common cast that also uses less parentheses, e.g.:
```go
int(7.5 + 0.37*compression - 2e-4*compression*compression)
```

Rather than currently:
```go
(int)(7.5 + 0.37*compression - 2e-4*compression*compression)
```

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+	d.add(value, 1.0)
+}
+
+func (d *tDigest) Quantile(q float64) float64 {
+	// compress the centroids first
+	d.compress()
+
+	var (
+		targetWeight = q * d.mergedWeight
+		curWeight    = 0.0
```
robskillington: `curWeight` to match `currWeight`?

robskillington: i.e. some places we use `curThing` and other places we use `currThing` perhaps standardize on one?

quantile/tdigest/tdigest.go
xichen2020: yeah can change that

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+
+	sort.Sort(centroidsByMeanAsc(d.unmerged))
+
+	var (
+		totalWeight = d.mergedWeight + d.unmergedWeight
+		currWeight  = 0.0
+		currIndex   = 0.0
+		merged      = d.merged
+		unmerged    = d.unmerged
+		buffer      = d.unmerged[:0]
```
xichen2020: we could, but it means each tdigest will have another buffer that's as big as the sum of the sizes of the two buffers, and we might have many tdigests. 

I could change it if the in-place merging code is fairly unreadable.

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+	)
+
+	d.merged = d.merged[:0]
+	for len(merged) > 0 || len(buffer) > 0 || len(unmerged) > 0 {
```
xichen2020: sure

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+
+		if currUnmerged.Mean < currMerged.Mean {
+			// Save the first centroid in case it'll be overwritten
+			if len(merged) > 0 {
+				buffer = d.appendCentroid(buffer, merged[0])
+				merged = merged[1:]
+			}
+			unmerged = unmerged[1:]
+			currIndex = d.mergeCentroid(currIndex, currWeight, totalWeight, currUnmerged)
+			currWeight += currUnmerged.Weight
```
xichen2020: sure

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// nextIndex estimates the index of the next centroid
+func (d *tDigest) nextIndex(quantile float64) float64 {
+	return d.compression * (math.Asin(2*quantile-1)/math.Pi + 0.5)
+}
+
+// add adds a weighted value
+func (d *tDigest) add(value float64, weight float64) {
+	if len(d.unmerged) == d.unmergedCapacity {
```
xichen2020: Don't think we can ever get to `>=` though without getting through `==` first

quantile/tdigest/tdigest.go
```diff
@@ -0,0 +1,285 @@
...
+// d.merged is guaranteed to have at least one centroid when upperBound is called.
+func (d *tDigest) upperBound(index int) float64 {
+	if index == len(d.merged)-1 {
+		return d.maxValue
+	}
+	return (d.merged[index].Mean + d.merged[index+1].Mean) / 2.0
+}
+
+func (d *tDigest) appendCentroid(centroids []Centroid, c Centroid) []Centroid {
+	if len(centroids) == cap(centroids) {
```
xichen2020: From `compress`, it only resets `buffer` pointer when `len(merged) == 0`, and once that is true, we'll never call `appendCentroid` afterwards, so this should be a non issue.

glide.yaml
xichen2020: Done

quantile/tdigest/tdigest.go
xichen2020: Was just using the same formula in tdunning's implementation, made the buffer size bigger now

quantile/tdigest/tdigest.go
xichen2020: It now uses an extra buffer for merging centroids. The code should be more readable now.

quantile/tdigest/tdigest_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/m3db/m3aggregator/pool"
```
robskillington: An extra new line below this perhaps?

quantile/tdigest/tdigest_test.go
```diff
@@ -0,0 +1,230 @@
...
+// THE SOFTWARE.
+
+package tdigest
+
+import (
+	"math"
+	"math/rand"
+	"testing"
+
+	"github.com/m3db/m3x/pool"
```
robskillington: nit: Endline below this import perhaps?

quantile/tdigest/tdigest_test.go
xichen2020: Yeah wil do

PR #3, opened 2016-12-13
MessagePack-based metrics encoder and decoder

xichen2020:
cc @robskillington @martin-mao @kobolog 

This PR adds encoder and decoder for serializing and deserializing batch of metrics using the MessagePack protocol.

### DIFF
--- a/protocol/msgpack/encoder.go
+++ b/protocol/msgpack/encoder.go
@@ -1,0 +1,211 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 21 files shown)


protocol/msgpack/encoder.go
```diff
@@ -0,0 +1,211 @@
...
+	if batchSize <= 0 {
+		return fmt.Errorf("batch size must be positive but is instead %d", batchSize)
+	}
+	enc.encodeBatchSize(batchSize)
+	enc.encodeVersion(supportedVersion)
+	return enc.err
+}
+
+func (enc *batchEncoder) EncodeCounter(c metric.Counter, p policy.VersionedPolicies) error {
+	enc.encodeType(metric.CounterType)
```
kobolog: AFAIK, the usual way to encode related data in MessagePack is to present it as tuples. For example, a counter with policies would be a tuple of `(metric.CounterType, "my.amazing.counter", 42, (PolicyA, ..., PolicyN))`. So, essentially, each "message" on the wire is represented as a set of nested tuples. Thoughts?

xichen2020: Yeah this might be the c++ way, but unfortunately the go version of the API (https://godoc.org/gopkg.in/vmihailenco/msgpack.v2#Encoder.Encode) converts each argument passed in to an interface{}, and then try to figure out what the type is (sometimes via reflection too), and then encodes accordingly based on the type. That's a big hit on perf based on my benchmark, and adds GC overhead too due to the extra allocations we need to do for the interface conversion, so for the simple types I'm trying to serialize and deserialize, I'm resorting to the lower-level APIs that don't require this sort of interface conversions.

xichen2020: Also I'm moving this logic to m3metrics since the encoder/decoder are likely to be used from many other repos as well, so I'm gonna close this PR and open a new one instead.

kobolog: Okay, I understand the `interface{}` part – I'm talking about a little bit different thing here, speicfically about the way things are represented on the wire. So, essentially I mean that instead of calling `encoder.EncodeInt(CounterType); encoder.EncodeBytes("my.amazing.counter"); ...` I'd rather do `encoder.EncodeArrayLen(5)` first, to specify that it's a tuple of size 5 (or whatever the number of fields is), and then call all the other `EncodeWhatever()` calls. So on the wire it's gonna be a 5-item tuple.

xichen2020: hm, what does that buy us? Is this for versioning?

kobolog: It's sort of like wire incapsulation – you will have clear object boundaries. This helps with versioning, schema evolution, etc. because when you unpack such an object, you read the whole thing from the wire, unpack the tuple and then just access the fields that you know about, i.e. adding new fields on sending side won't require any code changes on the receiving side.

xichen2020: Talked offline, seems reasonable for backward-compatible changes (e.g., add a field at the end etc). Gonna try to get this one in first, and then make that change in a separate PR.

xichen2020:
Moving this PR to m3metrics, closing this one

PR #4, opened 2017-01-10
Aggregation server

xichen2020:
cc @robskillington @martin-mao @kobolog @cw9 @prateek 

This PR adds the aggregation server implementation, along with unit tests, integration tests, and standalone service main functions.


aggregator/list.go
```diff
@@ -0,0 +1,340 @@
...
+type MetricLists struct {
+	sync.RWMutex
+
+	opts            Options
+	newMetricListFn newMetricListFn
+	closed          bool
+	lists           map[time.Duration]*MetricList
+}
+
+func newMetricLists(opts Options) *MetricLists {
```
cw9: do all the metriclists need to share the same timelock? should each metriclist have a separate timelock?

aggregation/quantile/cm/options.go
```diff
@@ -62,51 +64,66 @@ func NewOptions() Options {
 	floatsPool := NewFloatsPool(defaultBuckets, nil)
 	floatsPool.Init()
 
-	return options{
+	return &options{
```
xichen2020: More or less for consistency, since i'm also using a pointer for aggregator options. Passing pointers is cheaper than passing values, especially given how many fields the aggregator options have.

aggregation/types.go
```diff
@@ -0,0 +1,80 @@
...
+	sum int64 // Sum of the values received
+}
+
+// Gauge aggregates gauge values. Gauge APIs are thread-safe
+type Gauge struct {
+	value uint64 // latest value received
+}
+
+// Timer aggregates timer values. Timer APIs are not thread-safe
+type Timer struct {
```
xichen2020: Sure can do

aggregator/elem.go
```diff
@@ -0,0 +1,366 @@
...
+	value float64,
+	policy policy.Policy,
+)
+
+type newMetricElemFn func() metricElem
+
+// metricElem is the common interface for metric elements
+type metricElem interface {
+	// ResetSetData resets the counter and sets data
+	ResetSetData(id metric.ID, policy policy.Policy)
```
xichen2020: It not only resets its internal state, but also sets the initial data, hence `ResetSetData`

aggregator/elem.go
```diff
@@ -0,0 +1,366 @@
...
+
+// GaugeElem is the gauge element
+type GaugeElem struct {
+	elemBase
+
+	values map[time.Time]*aggregation.Gauge
+}
+
+// ResetSetData resets the counter and sets data
+func (e *elemBase) ResetSetData(id metric.ID, policy policy.Policy) {
```
xichen2020: We call `ResetSetData` as part of initializing the elem before putting it to use, so shouldn't need to lock

aggregator/entry.go
```diff
@@ -0,0 +1,242 @@
...
+
+	e.addMetricWithLock(currTime, mu)
+	e.Unlock()
+
+	return nil
+}
+
+// MaybeExpire attempts to expire the entry, returning true
+// if the entry has been expired and false otherwise
+func (e *Entry) MaybeExpire(now time.Time) bool {
```
xichen2020: I'm not sure, we are not really trying to expire the ones that shouldn't expire. Instead, we are just testing each one to see if we we should expire and expire if needed.

aggregator/entry.go
```diff
@@ -0,0 +1,242 @@
...
+// and we've surpassed the cutover, we should update the policies
+func (e *Entry) shouldUpdatePoliciesWithLock(
+	currTime time.Time,
+	version int,
+	cutover time.Time,
+) bool {
+	return e.version == policy.InitPolicyVersion || (e.version < version && !currTime.Before(cutover))
+}
+
+func (e *Entry) updatePoliciesWithLock(
```
xichen2020: Good catch thanks, will catch it in a test

aggregator/list.go
xichen2020: Yeah possibly, although it's easier to program with a shared lock, since the first thing the aggregator does when adding an entry is to acquire the read lock, if we make it per-list, it'll need to figure out which ones from which lists it needs to RLock, which seems messy.

The critical section within a write lock is really just getting a time value, which should be done instantly, so I think it should be fine. Can optimize further if this becomes the point of contention.

aggregation/quantile/cm/options.go
cw9: sure, sounds good

aggregator/elem.go
cw9: kk, maybe ResetWith() is slightly better?

aggregator/elem.go
cw9: sg

aggregator/list.go
cw9: yeah make sense, probably not so critical given the current usage

aggregator/elem.go
xichen2020: We use `ResetSetData` in other repos too, so it might be slightly more consistent, i'm not feeling strongly either way

cw9: sure, I don't feel strongly either way as well

server/packet/queue.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// Len returns the number of items in the queue
+func (q *Queue) Len() int { return len(q.queue) }
+
+// Enqueue pushes a packet into the queue
+func (q *Queue) Enqueue(p Packet) error {
+	callStart := q.nowFn()
+	for atomic.LoadInt32(&q.closed) == 0 {
```
prateek: The current way of doing things can starve earlier calls if the scheduler interleaves a lot of Enqueue calls, is that something to be concerned about?

aggregation/timer.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Stdev returns the standard deviation timer value
+func (t *Timer) Stdev() float64 {
+	num := float64(t.count)*t.sumSq - math.Pow(t.sum, 2)
+	div := t.count * (t.count - 1)
+	if div == 0 {
+		return 0.0
+	}
+	return math.Sqrt(num / float64(div))
```
robskillington: I thought this should be `sqrt( 1/(N-1) * summation( pow(x[i] - mean, 2) ) )`?  

I haven't looked hard at this but I can't see how this evaluates to that?  After finding another golang implementation it's a lot easier to map that equation to their implementation:
https://github.com/montanaflynn/stats/blob/60dcacf48f43d6dd654d0ed94120ff5806c5ca5c/variance.go#L6

It seems yours is:
`sqrt( 1/(N*(N-1)) * N * summation( pow(x[i] - totalsum, 2) ) )`

It looks equivalent but I worry if N is shaped such that the divisor could be zero and that it may not equate to the prior definition.

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+	ResetSetData(id metric.ID, policy policy.Policy)
+
+	// AddMetric adds a new metric value
+	// TODO(xichen): a value union would suffice here
+	AddMetric(timestamp time.Time, mu unaggregated.MetricUnion) error
+
+	// ReadAndDiscard processes values before a given time and discards
+	// them afterwards, returning whether the element can be collected
+	// after discarding the values
+	ReadAndDiscard(earlierThan time.Time, fn aggMetricFn) bool
```
robskillington: Discard implies also closing, (well at least in M3DB), may a better name be `Consume(...)`?

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+// GaugeElem is the gauge element
+type GaugeElem struct {
+	elemBase
+
+	values []timedGauge // aggregated gauges sorted by time in ascending order
+}
+
+// ResetSetData resets the counter and sets data
+func (e *elemBase) ResetSetData(id metric.ID, policy policy.Policy) {
+	e.id = id
```
robskillington: Missing locking here?

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+		if e.values[idx].timeNs >= earlierThanNs {
+			break
+		}
+		endAt := time.Unix(0, e.values[idx].timeNs).Add(e.policy.Resolution.Window)
+		e.processValue(endAt, e.values[idx].counter, fn)
+		idx++
+	}
+	if idx > 0 {
+		// Shift remaining values to the left and shrink the values slice
+		n := copy(e.values[0:], e.values[idx:])
```
robskillington: Is this safe in terms of overlapping?  i.e. if this copy was right to left from dst to source then would be an issue.  If we're sure and its documented its left to right and won't clobber then obviously ok.

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+		}
+		fn(fullTimerPrefix, e.id, quantileSuffixes[idx], timestamp, v, e.policy)
+	}
+}
+
+// NewGaugeElem creates a new gauge element
+func NewGaugeElem(id metric.ID, policy policy.Policy, opts Options) *GaugeElem {
+	return &GaugeElem{
+		elemBase: elemBase{opts: opts, id: id, policy: policy},
+		values:   make([]timedGauge, 0, 2), // in most cases values will have two entries
```
robskillington: Perhaps make this 2 a const?

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+		return left
+	}
+	// Otherwise we need to insert a new item
+	e.values = append(e.values, emptyTimedGauge)
+	// NB(xichen): it is ok if the source and the destination overlap
+	copy(e.values[left+1:numValues+1], e.values[left:numValues])
+	e.values[left] = timedGauge{
+		timeNs: alignedStart,
+		gauge:  aggregation.NewGauge(),
+	}
```
robskillington: Really kind of sucks we have to have this bucket searching implemented 3 times, there really is no way to generalize it or use a struct that is a union of the three types and we just use different field based on usecase?

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+// Close closes the gauge element
+func (e *GaugeElem) Close() {
+	e.Lock()
+	if e.closed {
+		e.Unlock()
+		return
+	}
+	e.closed = true
+	e.id = nil
+	e.values = e.values[:0]
```
robskillington: No need to nil the gauge values out to release any refs?

aggregator/elem.go
```diff
@@ -0,0 +1,508 @@
...
+// Close closes the counter element
+func (e *CounterElem) Close() {
+	e.Lock()
+	if e.closed {
+		e.Unlock()
+		return
+	}
+	e.closed = true
+	e.id = nil
+	e.values = e.values[:0]
```
robskillington: No need to nil the counter values out to release any refs?

aggregator/entry.go
```diff
@@ -0,0 +1,243 @@
...
+
+// IncWriter increases the writer count
+func (e *Entry) IncWriter() { atomic.AddInt32(&e.numWriters, 1) }
+
+// DecWriter decreases the writer count
+func (e *Entry) DecWriter() { atomic.AddInt32(&e.numWriters, -1) }
+
+// ResetSetData resets the entry and sets initial data
+func (e *Entry) ResetSetData(lists *MetricLists) {
+	e.closed = false
```
robskillington: No locking here?

aggregator/map.go
```diff
@@ -0,0 +1,193 @@
...
+func (m *metricMap) DeleteExpired(target time.Duration) {
+	now := m.nowFn()
+
+	// Determine batch size
+	m.RLock()
+	elemsLen := m.entryList.Len()
+	batchSize := int(math.Max(1.0, math.Ceil(m.batchPercent*float64(elemsLen))))
+	numBatches := int(math.Ceil(float64(elemsLen) / float64(batchSize)))
+	var targetPerBatch time.Duration
+	if numBatches > 0 {
```
martin-mao: Why not early exit for when elemsLen is 0 instead of doing all these calculations only to have numBatches to be 0.

aggregator/map.go
```diff
@@ -0,0 +1,193 @@
...
+	m.RUnlock()
+
+	// NB(xichen): if this runs frequently enough, stash the expired buffer
+	// in the map object and reuse the buffer
+	expired := make([]hashedEntry, 0, batchSize)
+	batchIdx := 0
+	for currElem != nil {
+		m.RLock()
+		for numChecked := 0; numChecked < batchSize && currElem != nil; numChecked++ {
+			nextElem := currElem.Next()
```
martin-mao: Since we're only doing a Rlock around here, can the entryList change in between the beginning of this func and halfway in between? Esp since there are long waits. If so (ie elems can be added), will that potentially much around with the batchSize?

integration/client.go
```diff
@@ -0,0 +1,128 @@
...
+		encoder:   msgpack.NewUnaggregatedEncoder(msgpack.NewPooledBufferedEncoder(nil)),
+	}
+}
+
+func (c *client) connect() error {
+	conn, err := net.DialTimeout("tcp", c.address, c.connectTimeout)
+	if err != nil {
+		return err
+	}
+	c.conn = conn
```
martin-mao: What if this is called twice? Need to check the current c.conn and close it?

integration/client.go
```diff
@@ -0,0 +1,128 @@
...
+		err = fmt.Errorf("unrecognized metric type %v", mu.Type)
+	}
+	if err != nil {
+		encoder.Buffer.Truncate(sizeBefore)
+		c.encoder.Reset(encoder)
+		return err
+	}
+	sizeAfter := encoder.Buffer.Len()
+	// If the buffer size is not big enough, do nothing
+	if sizeAfter < c.batchSize {
```
martin-mao: There is no timer here that flushes periodically. What happens if the client only writes once? Are they expected to flush themselves periodically? If so, maybe we should call this out explicitly, that this client will flush once it reaches batchSize, but won't do it periodically. 

Although, I imagine most users will want built-in periodic flush support.

integration/data.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	testCounterVal        = int64(123)
+	testBatchTimerVals    = []float64{1.5, 2.5, 3.5, 4.5, 5.5}
+	testGaugeVal          = 456.789
+	testVersionedPolicies = policy.VersionedPolicies{
+		Version: 1,
+		Policies: []policy.Policy{
```
martin-mao: Do we have any benchmark tests here? It would be good to know what the cost of applying each policy is and what the upper limit is both in terms of number of policies and number of time series per policy.

integration/client.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	// Otherwise we get a new buffer and copy the bytes exceeding the max
+	// flush size to it, swap the new buffer with the old one, and flush out
+	// the old buffer
+	encoder2 := msgpack.NewPooledBufferedEncoder(nil)
+	data := encoder.Bytes()
+	encoder2.Buffer.Write(data[sizeBefore:sizeAfter])
+	c.encoder.Reset(encoder2)
+	encoder.Buffer.Truncate(sizeBefore)
+	_, err = c.conn.Write(encoder.Bytes())
```
martin-mao: Can we set a write timeout that's passed in via options. We don't want to use default vals here for when server side dies in certain ways.

services/m3aggregator/main/main.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/m3db/m3metrics/policy"
+	"github.com/m3db/m3x/log"
+)
+
+const (
+	gracefulShutdownTimeout = 10 * time.Second
+)
+
+var (
+	msgpackAddrArg = flag.String("msgpackAddr", "0.0.0.0:6000", "msgpack server address")
```
martin-mao: Add TODO here to move these to options?

aggregation/timer.go
xichen2020: This is ported over from statsite implementation:
https://github.com/statsite/statsite/blob/master/src/timer.c#L95

Given that it's working fine for us so far, I'd rather not touch the math part.

aggregator/entry.go
xichen2020: Not necessary since we only call it when it's out of the pool before it can be used

aggregator/elem.go
xichen2020: Sure can change it

aggregator/elem.go
xichen2020: Don't need a lock here because the method is called before it's put to use

aggregator/elem.go
xichen2020: Yes it's safe to do.
```
The source and destination may overlap.
```

aggregator/elem.go
xichen2020: No need, gauge is a struct with no references/pointers

aggregator/elem.go
xichen2020: No need, counter is a struct with no references/pointers

aggregator/elem.go
xichen2020: Sure thing

aggregator/elem.go
xichen2020: Agreed it sucks but unfortunately go's lack of generics makes it very difficult to do this cleanly. I tried various approaches here including the one you mentioned, however the values will end up three times as big (or we need to allocate things on the heap using pointers), which is not good because these are high-cardinality objects, and there are be type checks (e.g., whether this is a counter/gauge/timer) and conditional operations littered all over the place, which makes the code very messy and hard to read. Therefore I fell back to the copy-paste approach, at least it's easy to tell what each method is doing this way without additional memory/GC overhead.

aggregation/counter.go
```diff
@@ -0,0 +1,39 @@
...
+// Counter aggregates counter values
+type Counter struct {
+	sum int64 // Sum of the values received
+}
+
+// NewCounter creates a new counter
+func NewCounter() Counter { return emptyCounter }
+
+// Add adds a counter value
+func (c *Counter) Add(value int64) { c.sum += value }
```
xichen2020: Correct, intentionally made it so because it's already protected by the callers, so no need to add the extra overhead of atomics

aggregation/gauge.go
```diff
@@ -0,0 +1,39 @@
...
+// Gauge aggregates gauge values
+type Gauge struct {
+	value float64 // latest value received
+}
+
+// NewGauge creates a new gauge
+func NewGauge() Gauge { return emptyGauge }
+
+// Add adds a gauge value
+func (g *Gauge) Add(value float64) { g.value = value }
```
xichen2020: Same explanation, it's already protected by the callers. Can change it to `Set`

aggregator/list.go
```diff
@@ -0,0 +1,342 @@
...
+		go l.tick()
+	}
+
+	return l
+}
+
+// Close closes the list
+func (l *MetricList) Close() {
+	l.Lock()
+	if l.closed {
```
xichen2020: Yeah basically I'm using a bool if there is already a `RWMutex` in the object to save the overhead of atomics, and a uint32 otherwise

aggregator/list.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+}
+
+func (l *MetricList) tickInternal() {
+	// NB(xichen): it is important to determine ticking start time within the time lock
+	// because this ensures all the actions before `start` have completed if those actions
+	// are protected by the same read lock
+	l.timeLock.Lock()
+	start := l.nowFn()
+	alignedStart := start.Truncate(l.resolution)
```
xichen2020: Sure thing

aggregator/list.go
```diff
@@ -0,0 +1,342 @@
...
+	if sizeAfter < l.maxFlushSize {
+		return
+	}
+	// Otherwise we get a new buffer and copy the bytes exceeding the max
+	// flush size to it, swap the new buffer with the old one, and flush out
+	// the old buffer
+	encoder2 := l.encoderPool.Get()
+	data := encoder.Bytes()
+	encoder2.Buffer.Write(data[sizeBefore:sizeAfter])
+	l.encoder.Reset(encoder2)
```
xichen2020: The encoder will be returned to pool when it's closed, which will be done by the goroutine that consumes the bytes in the encoder buffer

aggregator/map.go
xichen2020: Sure can add the shortcut logic

aggregator/map.go
xichen2020: elements can only be added to the end of the list after acquiring a write lock. As long as the batch size is positive and close enough to the optimal batch size, we'll eventually get through the list.

integration/client.go
xichen2020: This is only used in integration tests so I'm not too worried

integration/client.go
xichen2020: As mentioned above, the client is only used for integration testing, the real aggregation server will flush metrics out periodically

integration/client.go
xichen2020: Not too concerned since this is only for integration testing

integration/data.go
xichen2020: Sure can add benchmark tests, although that'll heavily depends on cache locality and probably should be in m3metrics repo instead. Can open up a followup diff for that.

server/packet/queue.go
xichen2020: Not too concerned about that possibility here, the queue should be properly sized to buffer metrics for certain amount of time, and if the queue does become full, we'll discard oldest metrics to make room in the queue

services/m3aggregator/main/main.go
xichen2020: These are used to stand up an example server quickly. In production we'll have a different main file internally that reads from config files.

aggregator/aggregator.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (agg *aggregator) tick() {
+	defer agg.wgTick.Done()
+
+	for {
+		select {
+		case <-agg.doneCh:
+			return
+		default:
+			agg.tickInternal()
```
kobolog: Is there any specific reason to have a separate `tickInternal()` function? If not, I'd rather combine them because having the `select` statement split into two parts is confusing.

aggregator/entry.go
```diff
@@ -0,0 +1,243 @@
...
+	e.lists = lists
+	e.version = policy.InitPolicyVersion
+	e.numWriters = 0
+	e.recordLastAccessed(e.opts.ClockOptions().NowFn()())
+}
+
+// AddMetricWithPolicies adds a metric along with applicable policies
+func (e *Entry) AddMetricWithPolicies(
```
kobolog: JFYI: I found it somewhat more elegant to write this kind of code (RLock -> Check -> RUnlock -> Lock -> Check -> Action -> Unlock) by using `sync.Locker` objects, specifically by factoring out the common part into a function that accepts a `sync.Locker` to avoid repetition.

aggregator/entry.go
```diff
@@ -0,0 +1,243 @@
...
+		e.Unlock()
+		return
+	}
+	e.closed = true
+	for p, agg := range e.aggregations {
+		switch elem := agg.Value.(type) {
+		case metricElem:
+			elem.MarkAsTombstoned()
+		default:
+			// We should never get here
```
kobolog: If we should never get here, then why spend CPU cycles on that type assertion?

aggregator/list.go
```diff
@@ -0,0 +1,343 @@
...
+// MetricList stores aggregated metrics at a given resolution
+// and flushes aggregations periodically
+// TODO(xichen): add metrics
+type MetricList struct {
+	sync.RWMutex
+
+	opts         Options
+	nowFn        clock.NowFn
+	log          xlog.Logger
+	timeLock     *sync.RWMutex
```
kobolog: As far as I understand Go's memory model, having a pointer field inside a structure is an extra pointer on the heap, while a pointer on the stack set to an address of some non-pointer structure's field is not, while still allowing pointer-receiver function calls.

aggregator/aggregator.go
xichen2020: This is mostly for testing purposes so we can call `tickInternal` to force a tick in unit tests

aggregator/entry.go
xichen2020: I see, so you can pass in  `RLocker` first and `Locker` afterwards, makes sense.

Thankfully this is only used in one place for `Entry` so it;s not too bad, but it's a good tip!

aggregator/entry.go
xichen2020: Fair enough, was trying to make the error message a bit nicer but yeah can get rid of it

aggregator/list.go
xichen2020: Yeah I cache it here so that we don't have to call `opts.TimeLock()` on every tick, but I think either way is probably okay because the number of metric lists is small, and we don't tick that often (on the order of seconds)

aggregator/list.go
```diff
@@ -0,0 +1,343 @@
...
+	errListClosed  = errors.New("metric list is closed")
+	errListsClosed = errors.New("metric lists are closed")
+)
+
+type encodeFn func(mp aggregated.ChunkedMetricWithPolicy) error
+
+// MetricList stores aggregated metrics at a given resolution
+// and flushes aggregations periodically
+// TODO(xichen): add metrics
+type MetricList struct {
```
robskillington: Is `MetricList` used outside of this package?  If not does it need to be exported?

aggregator/list.go
```diff
@@ -0,0 +1,343 @@
...
+	encoder.Buffer.Truncate(sizeBefore)
+	if err := l.flushFn(encoder); err != nil {
+		l.log.Errorf("flushing metrics error: %v", err)
+	}
+}
+
+type newMetricListFn func(resolution time.Duration, opts Options) *MetricList
+
+// MetricLists contains all the metric lists
+type MetricLists struct {
```
robskillington: Is `MetricLists` used outside of this package?  If not does it need to be exported?

aggregator/entry.go
```diff
@@ -0,0 +1,237 @@
...
+
+var (
+	errEntryClosed = errors.New("entry is closed")
+)
+
+// Entry stores metadata about current policies and aggregations
+// for a metric. Note since the metric namespace is part of the
+// metric ID, we only need to keep track of the policy version
+// to detect policy changes
+type Entry struct {
```
robskillington: Is `Entry` used outside of this package?  If not does it need to be exported?

aggregation/timer.go
```diff
@@ -0,0 +1,107 @@
...
+package aggregation
+
+import (
+	"math"
+
+	"github.com/m3db/m3aggregator/aggregation/quantile/cm"
+)
+
+// Timer aggregates timer values. Timer APIs are not thread-safe
+type Timer struct {
```
robskillington: Is `Timer` used outside of this package?  If not does it need to be exported?

aggregator/elem.go
robskillington: Sure thing.

aggregator/elem.go
robskillington: Sure thing.

aggregator/options.go
```diff
@@ -0,0 +1,515 @@
...
+	defaultTimerSumSuffix         = []byte(".sum")
+	defaultTimerSumSqSuffix       = []byte(".sum_sq")
+	defaultTimerMeanSuffix        = []byte(".mean")
+	defaultTimerLowerSuffix       = []byte(".lower")
+	defaultTimerUpperSuffix       = []byte(".upper")
+	defaultTimerCountSuffix       = []byte(".count")
+	defaultTimerStdevSuffix       = []byte(".stdev")
+	defaultTimerMedianSuffix      = []byte(".median")
+	defaultGaugePrefix            = []byte("gauges.")
+	defaultMinFlushInterval       = 5 * time.Second
```
robskillington: This should probably always be the `policy resolution / 2` right?  Perhaps each list should have a different one that can default to this but should be at the very least `policy resolution / 2`?

Otherwise how does the 1s lists get flushed fast enough.

aggregator/list.go
```diff
@@ -0,0 +1,343 @@
...
+
+func (l *MetricList) tickInternal() {
+	// NB(xichen): it is important to determine ticking start time within the time lock
+	// because this ensures all the actions before `start` have completed if those actions
+	// are protected by the same read lock
+	l.timeLock.Lock()
+	start := l.nowFn()
+	resolution := l.resolution
+	l.timeLock.Unlock()
+	alignedStart := start.Truncate(resolution)
```
robskillington: If for some reason the tick doesn't wake up in time for this to be the very next interval then we skip one it seems, is that correct?  

Just want to get a handle on the exact architecture here.

aggregator/list.go
```diff
@@ -0,0 +1,343 @@
...
+		case metricElem:
+			// If the element is eligible for collection after the values are
+			// processed, close it and reset the value to nil
+			if elem.Consume(alignedStart, l.aggMetricFn) {
+				elem.Close()
+				e.Value = nil
+				l.toCollect = append(l.toCollect, e)
+			}
+		default:
+			panic(fmt.Errorf("unknown metric element type: %T", elem))
```
robskillington: Seems some places we check the type and panic if not and others we just cast and call directly against the cast.  Perhaps just do one or the other?  I feel we probably should just be casting and calling on it rather than checking type.  We'll panic anyway just in a slightly different way and its less code overall.

aggregator/entry.go
xichen2020: Yes it's used as the return type from the `Get` method of the entry pool, which is exported.

server/msgpack/server.go
```diff
@@ -0,0 +1,240 @@
...
+	}()
+
+	it := s.iteratorPool.Get()
+	it.Reset(bufio.NewReader(conn))
+	defer it.Close()
+
+	// Iterate over the incoming metrics stream and queue up metrics
+	for it.Next() {
+		metric, policies := it.Value()
+		if err := s.queue.Enqueue(packet.Packet{Metric: metric, Policies: policies}); err != nil {
```
xichen2020: I feel the packet queue and the packet processor provide a good level of isolation between the producer (packets coming in) and the consumer (the aggregation layer).

By enqueuing the packets to the queue, this provides load shedding when the queue is overloaded. Additionally, when the queue is overwhelmed and needs to drop packets, the semantics is such that it'll drop the oldest packets and make room for the recent packets, which is important because the recent data are more valuable for monitoring and alerting.

The processor provides the ability to limit the parallelism for packet processing, so that the number of package processing goroutines is no more than the number of cores, avoiding the high context switching overhead when we have one goroutine per connection. Additionally, if one goroutine is temporarily slow when processing packets, other goroutines can pick up the slack without blocking packets coming in from any particular connection.

If the queue turns out to be a hotspot during testing, it doesn't require much change to connect the server directly to the aggregator, but for now I'd rather keep them for the reasons mentioned above unless I have more production data to show otherwise.

server/packet/processor.go
```diff
@@ -0,0 +1,123 @@
...
+		log:        instrumentOpts.Logger(),
+		metrics:    newProcessorMetrics(scope, samplingRate),
+	}
+
+	// Start the workers to drain the queue
+	p.wgWorkers.Add(numWorkers)
+	p.workers = xsync.NewWorkerPool(numWorkers)
+	p.workers.Init()
+	for i := 0; i < numWorkers; i++ {
+		p.workers.Go(p.drain)
```
xichen2020: Yeah sounds good, will change to use goroutines

server/packet/processor.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return p
+}
+
+// Close closes the processor. It's safe to call close more than once.
+// All but the first call are no-ops.
+func (p *Processor) Close() {
```
xichen2020: I feel it's more idiomatic to call it `Close` since it means the processor is closing itself and once closed, will be unavailable for any further processing, so I prefer to keep it unless you feel strongly.

services/m3aggregator/processor/processor.go
```diff
@@ -0,0 +1,141 @@
...
+	queue     chan msgpack.BufferedEncoder
+	workers   xsync.WorkerPool
+	wgWorkers sync.WaitGroup
+	fn        MetricWithPolicyFn
+	log       xlog.Logger
+	iterOpts  msgpack.AggregatedIteratorOptions
+}
+
+// NewAggregatedMetricProcessor creates a new aggregated metric processor
+func NewAggregatedMetricProcessor(opts Options) *AggregatedMetricProcessor {
```
xichen2020: I'm not sure I like the idea of unbounded memory growth under high load. It feels better that we preallocate enough space for the channel based on the total memory capacity and the number of processes running on the server, and just have the process always running in the steady state without having its memory usage shrinking or expanding a ton. If a particular process is under load, that process should take measures to alleviate pressure on itself (e.g., by dropping old metrics in queue). If we allow unlimited memory growth without shedding load, a badly behaving process might use a ton of memory, negatively affecting the performance of its good neighbors on the same box. FWIW this is the same approach we adopt in shunts.

aggregator/elem.go
robskillington: Sounds good.

aggregator/elem.go
robskillington: Sounds good.

aggregator/entry.go
```diff
@@ -0,0 +1,243 @@
...
+// ShouldExpire returns whether the entry should expire
+func (e *Entry) ShouldExpire(now time.Time) bool {
+	e.RLock()
+	if e.closed {
+		e.RUnlock()
+		return false
+	}
+	// Only expire the entry if there are no active writers
+	// and it has reached its ttl since last accessed
+	shouldExpire := e.writerCount() == 0 && now.After(e.lastAccessed().Add(e.opts.EntryTTL()))
```
robskillington: Sounds good.

services/m3aggregator/main/main.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Creating the downstream processor
+	processorOpts := processor.NewOptions()
+	log := processorOpts.InstrumentOptions().Logger()
+	metricWithPolicyfn := func(metric aggregated.Metric, policy policy.Policy) error {
+		log.WithFields(
+			xlog.NewLogField("metric", metric.String()),
+			xlog.NewLogField("policy", policy.String()),
+		).Info("aggregated metric")
```
robskillington: Yeah, all I'm saying is make this take an interface rather than a function pointer.  How do you feel about that?

services/m3aggregator/main/main.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil
+	}
+	processorOpts = processorOpts.SetMetricWithPolicyFn(metricWithPolicyfn)
+	processor := processor.NewAggregatedMetricProcessor(processorOpts)
+
+	// Creating the aggregator
+	aggregatorOpts := aggregator.NewOptions().SetFlushFn(processor.Add)
```
robskillington: You can't get a reference to the raw bytes from the buffer?  At the very least can we wrap it as a different type.

e.g. pass instead a:
```go
type Buffer interface {
  io.Reader

  // Close the buffer
  Close()
}
```

That way you can pass the exact same struct here but the intention of how it should be used is clear.

services/m3aggregator/processor/processor.go
robskillington: Sure but now you have to size the queues to match the memory of the machine.  If you overshoot then the process always out of memories when you start it up and if you undershoot you start dropping metrics when you clearly could still make progress and buffer.

server/packet/processor.go
robskillington: Maybe make the processor own the queue itself and close the queue in here?  That seems less coupled and more idiomatic.

server/server.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/m3db/m3x/close"
+)
+
+// Server is a server capable of listening to incoming traffic and closing itself
+// when it's shut down
+type Server interface {
+	// ListenAndServe starts listening to new incoming connections and
+	// handles data from those connections
+	ListenAndServe() (xclose.SimpleCloser, error)
```
robskillington: Can you add a comment that specifically says it is non-blocking (which is different from Golang stdlib HTTP library).  Otherwise people will assume its blocking, similar to the stdlib HTTP library.

server/msgpack/server.go
robskillington: Can we benchmark this then?  

I don't see why we should funnel data from thousnads of asynchronous driven network IO goroutines into a single contended channel before coming out again on the otherside for not much reason other than for code isolation/structure.

aggregation/timer.go
robskillington: Missed that, cheers

aggregator/options.go
robskillington: Well in practice some teams want 1s deploy rollbacks eventually.  Maybe we can do something where each list is flushed at its own cycle?  That way we don't scan the really huge lists as frequently.

I'm probably ok living with this for now but we should open an issue to investigate how to provide expedited flushing for high resolution data perhaps?

aggregator/list.go
robskillington: Right yeah, I got my head around this after reading more code.  Sounds good 👍

services/m3aggregator/main/main.go
xichen2020: Yeah sure thing, can do that

services/m3aggregator/main/main.go
xichen2020: If I pass the raw bytes around, I'll have the same issue: when the raw bytes are returned to pool, the encoder needs to be returned to pool too.

I could try wrapping the buffer along with a callback that gets called when the buffer is returned to pool. Let me try that.

services/m3aggregator/processor/processor.go
xichen2020: Agreed, we'll need to size the queues properly, I feel that's ok because we can determine the size by load testing it before productionization and should hardly need to touch that again before hitting the load target. In general I prefer to run it in steady mode where the memory usage stays more or less constant regardless of load.

server/server.go
xichen2020: Sure thing

server/packet/processor.go
xichen2020: Yeah sounds good, will do that

aggregator/options.go
xichen2020: Each list is indeed flushed at its own cycle. The flush interval is the max of the minimum flush interval and the resolution, so with the default settings, the 1s list will flush every 5 seconds, and the 1 min list will flush every minute.

Sure can open an issue for expedited flushing. I imagine we can tag the metrics (maybe in the rules) somehow so that they live in their own 1s list with high flushing frequency.

aggregation/timer.go
robskillington: Sure thing.

glide.lock
```diff
@@ -8,12 +8,13 @@ imports:
 - name: github.com/facebookgo/clock
   version: 600d898af40aa09a7a93ecb9265d87b0504b6f03
 - name: github.com/golang/protobuf
-  version: bf531ff1a004f24ee53329dfd5ce0b41bfdc17df
+  version: 7390af9dcd3c33042ebaf2474a1724a83cf1a7e6
```
robskillington: Maybe we should pin protobuf so it doesn't update implicitly?

glide.lock
```diff
@@ -46,7 +46,7 @@ imports:
 - name: github.com/uber-go/tally
   version: 1f0f7171f312a8cff5f697fe5f4d45dbaec8c3ac
 - name: golang.org/x/net
-  version: 007e530097ad7f954752df63046b4036f98ba6a6
+  version: 3b993948b6f0e651ffb58ba135d8538a68b1cddf
```
robskillington: Maybe we should pin golang.org/x/net so it doesn't update implicitly?

aggregator/list.go
```diff
@@ -0,0 +1,339 @@
...
+	timeLock     *sync.RWMutex
+	maxFlushSize int
+	flushFn      FlushFn
+	encoderPool  msgpack.BufferedEncoderPool
+
+	resolution    time.Duration
+	flushInterval time.Duration
+	aggregations  *list.List
+	timer         *time.Timer
+	idBuf         bytes.Buffer
```
robskillington: Seems like `idBuf` is never used?

PR #7, opened 2017-04-17
Add config structs and config yaml to build service binary

xichen2020:
cc @martin-mao @cw9 @prateek  @jeromefroe  

This PR adds config structs and example yaml so we can build the service binary out of the open source repo.

### DIFF
--- a/services/m3aggregator/config/aggregator.go
+++ b/services/m3aggregator/config/aggregator.go
@@ -1,0 +1,353 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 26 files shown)


services/m3aggregator/config/aggregator.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Pool of gauge elements.
+	GaugeElemPool pool.ObjectPoolConfiguration `yaml:"gaugeElemPool"`
+
+	// Pool of buffered encoders.
+	BufferedEncoderPool pool.ObjectPoolConfiguration `yaml:"bufferedEncoderPool"`
+}
+
+// NewAggregatorOptions creates a new aggregator options.
+func (c *aggregatorConfiguration) NewAggregatorOptions(
```
cw9: if aggregatorConfiguration is already private, should NewAggregatorOptions be private too?

services/m3aggregator/config/aggregator.go
```diff
@@ -0,0 +1,353 @@
...
+// flushConfiguration contains configuration for flushing metrics.
+type flushConfiguration struct {
+	// Flushing function type.
+	Type string `yaml:"type" validate:"regexp=(^blackhole$|^logging$|^forward$)"`
+
+	// Forward flush configuration.
+	Forward *forwardFlushConfiguration `yaml:"forward"`
+}
+
+func (c *flushConfiguration) NewFlushFn() (aggregator.FlushFn, error) {
```
cw9: same

services/m3aggregator/config/config.go
```diff
@@ -0,0 +1,38 @@
...
+
+import "github.com/m3db/m3x/instrument"
+
+// Configuration contains top-level configuration.
+type Configuration struct {
+	// Metrics configuration.
+	Metrics instrument.MetricsConfiguration `yaml:"metrics"`
+
+	// Aggregator configuration.
+	Aggregator aggregatorConfiguration `yaml:"aggregator"`
```
cw9: if the root struct is public, these fields should probably be public as well

services/m3aggregator/config/aggregator.go
xichen2020: Generally I treat a type and its methods differently w.r.t. whether they should be exported. If a type is only used internally, I make it an unexported type. However, it doesn't necessarily mean all the methods of this type must be unexported. Instead, a method should be exported if it would be considered the type's public API, and unexported otherwise (another way of thinking about this is, if the type were exported, would you make the method exported?).

In this case, I made the ctor exported because if at some point we want to move `aggregatorConfiguration` to a standalone package, I would want to make this ctor exported because it's the type's public API, however the type itself is only used internall, hence it's unexported.

services/m3aggregator/config/config.go
xichen2020: It is fine (and probably better in this case) to export the fields for unmarshaling purposes but not their types since they are internal types and not use anywhere outside the package.

services/m3aggregator/config/aggregator.go
xichen2020: See reply above.

services/m3aggregator/config/aggregator.go
cw9: sounds good

services/m3aggregator/config/config.go
cw9: It's understandable that the field is exported for unmarshaling purpose, but if a user uses the Configuration struct, not sure what would the user get when calling configuration.Aggregator.

xichen2020: Yeah you can still use unexported types outside the package, except that you can only use exported methods on the type, which in our case worked out. That said,  I'll export the type to be consistent with conventions.

PR #8, opened 2017-04-18
Implement forward handler

xichen2020:
cc @martin-mao @cw9 @jeromefroe @prateek 

This PR adds the forward handler implementation to forward aggregated metrics to downstream services.


aggregator/handler/forward_options.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"time"
+
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/retry"
+)
+
+const (
+	defaultQueueSize           = 65536
```
martin-mao: Are we eventually going to do the queue size based on mem usage as opposed to some number?

services/m3aggregator/config/aggregator.go
```diff
@@ -281,40 +285,70 @@ func (c *streamConfiguration) NewStreamOptions(instrumentOpts instrument.Options
 	return opts, nil
 }
 
-// flushConfiguration contains configuration for flushing metrics.
-type flushConfiguration struct {
-	// Flushing function type.
+// flushHandlerConfiguration contains configuration for flushing metrics.
+type flushHandlerConfiguration struct {
+	// Flushing handler type.
 	Type string `yaml:"type" validate:"regexp=(^blackhole$|^logging$|^forward$)"`
```
martin-mao: This config seems weird, why restrict the handler types in both a const as well as yaml validation. Are we going to chain these handlers? Do we need all handlers or are any of them optional? Any reason why we shouldn't just define/build a handler chain in code as we need them? 

Also, the forward only applies to the Forward Handler - this seems odd too.

aggregator/handler/forward.go
```diff
@@ -0,0 +1,231 @@
...
+var (
+	errEmptyServerList = errors.New("empty server list")
+	errHandlerClosed   = errors.New("handler is closed")
+	errBufferQueueFull = errors.New("buffer queue is full")
+)
+
+type dropType int
+
+const (
+	dropCurrent dropType = iota
```
xichen2020: Sure, can make this `iota + 1`

aggregator/handler/forward.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (h *forwardHandler) enqueue(buffer msgpack.Buffer, dropType dropType) error {
+	h.RLock()
+	if h.closed {
+		h.RUnlock()
+		return errHandlerClosed
+	}
+	for {
+		select {
+		case h.bufCh <- buffer:
```
xichen2020: Yes, `Handle` may be called from multiple goroutines. We store aggregated metrics with the same resolution in the same list, and when the list flushes, it calls `Handle` to flush the encoded data buffer when the size reaches MTU. 

The case you mentioned could happen, but only if (1) the queue is full and not draining fast enough, and (2) two goroutines flush at the same time (e.g., the 10s list and 1min list start flushing concurrently), and (3) it takes more time to perform one read from the channel than encoding a full buffer (otherwise the dequeuing goroutine should've successfully enqueued its buffer), which is likely not the case. 

So I'd say it's relatively rare. And even if it happens, as long as the dequeuing goroutine can enqueue the buffer in two or three iterations, it is not an issue, since the buffered data are already labeled with the correct timestamp.

Also, we use similar logic in shunts in prod and the contention among goroutines is much higher since any connection (including carbon metrics from the shunt server each host and the statsite boxes) can enqueue metrics.

aggregator/handler/forward_options.go
xichen2020: We could fix the mem usage and calculate the queue size based on the limit and the buffer size yes, although considering a significant amount of memory will be used to store the aggregated metrics, not sure how useful it is to only limit the mem usage of the forwarding queue.

services/m3aggregator/config/aggregator.go
xichen2020: Users are supposed to choose one and only one handler type, which handles the aggregated metrics. There is no chaining involved. 

The reason why only the forward configuration in there because the other types of handlers (`logging` and `blackhole`) do not require any configuration. I could potentially change the configuration struct to something like
```
type flushHandlerConfiguration struct {
  Logging *loggingHandlerConfiguration `yaml:"logging"`
  Blackhole *blackholeHandlerConfiguration `yaml:"blackhole"`
  Forward *forwardHandlerConfiguration `yaml:"forward"`
}  
```
and return an error if no handler configuration or more than one handler configuration if specified, but the `loggingHandlerConfiguration` and `blackholeHandlerConfiguration` will be empty structs, so if you want to use the logging handler, in the yaml file you'll have something like
```
flush:
  logging: {}
```
Which is kinda weird looking as well.

aggregator/handler/forward.go
```diff
@@ -0,0 +1,231 @@
...
+	h.wg.Add(numServers)
+	for _, addr := range servers {
+		conn, err := h.tryConnect(addr)
+		if err != nil {
+			h.log.WithFields(
+				xlog.NewLogField("address", addr),
+				xlog.NewLogErrField(err),
+			).Error("error connecting to server")
+		}
+		go h.forwardToConn(addr, conn)
```
xichen2020: Yeah so this is so that when the servers start up, we already have most if not all connections set up and can start draining queues right away. Otherwise you might run into a situation where the servers start taking in writes, but we are still in the middle of trying to establish connections.

PR #10, opened 2017-04-21
Add logging configuration

xichen2020:
cc @cw9 @jeromefroe @martin-mao 

This PR adds logging configuration to the top-level config so we can configure logging levels and log files etc.

### DIFF
--- a/aggregator/handler/logging.go
+++ b/aggregator/handler/logging.go
@@ -24,12 +24,15 @@ import (
 	"github.com/m3db/m3aggregator/aggregator"
 	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3metrics/policy"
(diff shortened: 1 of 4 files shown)


aggregator/handler/logging.go
```diff
@@ -24,20 +24,30 @@ import (
...
 )
 
-var logger = xlog.NewLevelLogger(xlog.SimpleLogger, xlog.LogLevelInfo)
-
-func loggingHandler(metric aggregated.Metric, policy policy.Policy) error {
+func logMetricAndPolicy(
+	logger xlog.Logger,
+	metric aggregated.Metric,
+	policy policy.Policy,
+) error {
```
jeromefroe: Any reason we need to return an error here?

jeromefroe:
LGTM

PR #11, opened 2017-04-21
Fix flaky handler and server tests

xichen2020:
cc @jeromefroe @cw9 @martin-mao 

This PR fixes a few flaky forward handler tests due to parallel tests connecting to the same local test server.


jeromefroe:
LGTM

PR #12, opened 2017-04-23
Pull in m3metrics improvements for better perf and memory utilization

xichen2020:
cc @jeromefroe @cw9 @martin-mao 

This PR pulls in latest m3metrics PR (https://github.com/m3db/m3metrics/pull/35) and makes necessary changes to improve perf and memory utilization. The notable changes are:
* In normal cases when there are no policy changes, we reuse the same []byte in the unaggregated iterator for the metric id, therefore makes zero []byte allocation for ids. The only time when we need to create a new id is when there is a policy change, in which case we only make a copy if the id is not owned, and there are no existing reusable ids. 
* The policy slices are cached internally in the unaggregated iterator and are reused across metrics.
* The timer value []float64 slices are reused for non-large value slices (slices with no more than 1024 elements are considered non-large, the threshold is configurable). For large value slices, the unaggregated iterator will allocate it from the large floats pool, and the aggregator returns it to pool after adding all the values.

### DIFF
--- a/aggregator/mock/aggregator.go
+++ b/aggregator/mock/aggregator.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/m3db/m3metrics/metric"
 	"github.com/m3db/m3metrics/metric/unaggregated"
 	"github.com/m3db/m3metrics/policy"
 )
@@ -49,6 +50,10 @@ func (agg *mockAggregator) AddMetricWithPolicies(
 	mu unaggregated.MetricUnion,
 	policies policy.VersionedPolicies,
 ) error {
+	// Clone the metric and policies to ensure it cannot be mutated externally.
+	mu = cloneMetric(mu)
+	policies = clonePolicies(policies)
+
 	agg.Lock()
 
 	switch mu.Type {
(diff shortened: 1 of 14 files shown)


aggregator/mock/aggregator.go
```diff
@@ -107,3 +112,30 @@ func (agg *mockAggregator) Snapshot() SnapshotResult {
 
 	return result
 }
+
+func cloneMetric(m unaggregated.MetricUnion) unaggregated.MetricUnion {
+	mu := m
+	if !m.OwnsID {
```
jeromefroe: nit: Shouldn't we copy the `ID` here as well since we're cloning the metric to ensure it can't be mutated externally? Probably a moot point since we don't intend to modify ID's and this is just a mock but figured I would point it out.

jeromefroe:
Left one small nit but otherwise LGTM

aggregator/mock/aggregator.go
xichen2020: The meaning of `OwnsID` is that if it's true, whoever consumes the id takes the ownership of the id and it is guaranteed that the id cannot be mutated externally, and if it's false, whoever consumes the id needs to make a copy of the id (since the id is not owned by the union and can then be mutated by the owner).

PR #14, opened 2017-05-05
Use agg tier clock as source of truth to handle clock skews

xichen2020:
cc @cw9 @robskillington @jeromefroe @prateek 

This PR is part of the bigger change to handle clock skews between the collectors and the agg tier by having the collectors sending matching policies within the range of clock skews and the agg tier decide which policies should be applied based on its clock.

Still WIP, will update once the dependent changes in m3metrics are finished.


aggregator/entry.go
```diff
@@ -118,8 +172,8 @@ func (e *Entry) AddMetricWithPolicies(
 		return errEntryClosed
 	}
 
-	if e.shouldUpdatePoliciesWithLock(currTime, policies.Version, policies.Cutover) {
-		if err := e.updatePoliciesWithLock(mu.Type, mu.ID, mu.OwnsID, policies.Version, policies.Policies()); err != nil {
+	if e.shouldUpdatePoliciesWithLock(currTime, sp) {
```
jeromefroe: Do we need to acquire a write lock before we read the value of `e.closed`? Perhaps we can defer it to here before we call `shouldUpdatePoliciesWithLock`?

Also, perhaps we should distinguish the functions which need a read lock from those which need a write lock? For example, `shouldUpdatePoliciesWithLock` is called with both but `updatePoliciesWithLock` needs a lock since it actually updates `e`.

xichen2020: Yeah so this is doing double-checking in that we acquire a read lock first, and if the entry should be updated, we release the read lock and acquire the write lock, but in between lock promotion, `e.closed` might change, so we need to check it again.

aggregator/options.go
```diff
@@ -50,6 +51,10 @@ var (
 	defaultEntryTTL               = 24 * time.Hour
 	defaultEntryCheckInterval     = time.Hour
 	defaultEntryCheckBatchPercent = 0.01
+	defaultDefaultPolicies        = []policy.Policy{
+		policy.NewPolicy(10*time.Second, xtime.Second, 2*24*time.Hour),
+		policy.NewPolicy(time.Minute, xtime.Minute, 30*24*time.Hour),
```
cw9: 40 days?

config/m3aggregator.yaml
```diff
@@ -52,6 +52,9 @@ aggregator:
     type: logging
   entryTTL: 24h
   entryCheckInterval: 1h
+  defaultPolicies:
+    - 10s:2d
+    - 1m:30d
```
cw9: 40d?

services/m3aggregator/config/aggregator.go
```diff
@@ -104,6 +105,9 @@ type AggregatorConfiguration struct {
 	// EntryCheckBatchPercent determines the percentage of entries checked in a batch.
 	EntryCheckBatchPercent float64 `yaml:"entryCheckBatchPercent" validate:"min=0.0,max=1.0"`
 
+	// Default policies.
+	DefaultPolicies []policy.Policy `yaml:"defaultPolicies" validate:"nonzero"`
```
cw9: probably should not be validated nonzero here? should provide a defaultPolicies if no configuration is provided?

aggregator/entry.go
jeromefroe: Gotcha, so we need to grab to it to ensure `e` can't be closed.

aggregator/entry.go
```diff
@@ -173,52 +234,102 @@ func (e *Entry) TryExpire(now time.Time) bool {
...
+		e.metrics.emptyPoliciesList.Inc(1)
+		return policy.DefaultStagedPolicies, errEmptyPoliciesList
+	}
+	timeNanos := t.UnixNano()
+	for idx := len(pl) - 1; idx >= 0; idx-- {
+		if pl[idx].CutoverNanos <= timeNanos {
+			return pl[idx], nil
+		}
+	}
+	return pl[0], nil
```
xichen2020: This is the case where all policies are in the future, in which case we'll use the cached policy instead. If we use a default policy here, it may give us the wrong results.

aggregator/options.go
xichen2020: Sure, we'll have our own internal config though

config/m3aggregator.yaml
xichen2020: Sure, we'll provide this in our internal config though.

services/m3aggregator/config/aggregator.go
xichen2020: I'd prefer to force the users to specify a default set of policies.

PR #15, opened 2017-05-18
Support custom aggregation types

cw9:
This diff picks up the change made in m3metrics to support custom aggregation types, we decode the policies with compressed aggregation types and use as is, only decompress the aggregation types when changes were noticed in the policies to save CPU.

@xichen2020 @prateek @jeromefroe

### DIFF
--- a/aggregation/common.go
+++ b/aggregation/common.go
@@ -20,13 +20,26 @@
 
 package aggregation
 
-import "math"
+import (
+	"math"
 
-func stdev(count, sumSq, sum float64) float64 {
+	"github.com/m3db/m3metrics/policy"
+)
+
+func stdev(count int64, sumSq, sum float64) float64 {
 	div := count * (count - 1)
 	if div == 0 {
 		return 0.0
 	}
-	num := count*sumSq - math.Pow(sum, 2)
+	num := float64(count)*sumSq - sum*sum
 	return math.Sqrt(num / float64(div))
+}
+
+func isExpensive(aggTypes policy.AggregationTypes) bool {
+	for _, aggType := range aggTypes {
+		if aggType == policy.SumSq || aggType == policy.Stdev {
+			return true
+		}
+	}
+	return false
 }
--- a/aggregation/common.go
+++ b/aggregation/common.go
@@ -18,49 +18,15 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package handler
+package aggregation
 
-import (
-	"io"
+import "math"
 
-	"github.com/m3db/m3aggregator/aggregator"
-	"github.com/m3db/m3metrics/metric/aggregated"
-	"github.com/m3db/m3metrics/policy"
-	"github.com/m3db/m3metrics/protocol/msgpack"
-)
-
-// HandleFunc handles an aggregated metric alongside the policy.
-type HandleFunc func(metric aggregated.Metric, policy policy.StoragePolicy) error
-
-type decodingHandler struct {
-	handle HandleFunc
-}
-
-// NewDecodingHandler creates a new decoding handler with a custom handle function.
-func NewDecodingHandler(handle HandleFunc) aggregator.Handler {
-	return decodingHandler{handle: handle}
-}
-
-func (h decodingHandler) Handle(buffer msgpack.Buffer) error {
-	defer buffer.Close()
-
-	iter := msgpack.NewAggregatedIterator(buffer.Buffer(), msgpack.NewAggregatedIteratorOptions())
-	defer iter.Close()
-
-	for iter.Next() {
-		rawMetric, sp := iter.Value()
-		metric, err := rawMetric.Metric()
-		if err != nil {
-			return err
-		}
-		if err := h.handle(metric, sp); err != nil {
-			return err
-		}
+func stdev(count, sumSq, sum float64) float64 {
+	div := count * (count - 1)
+	if div == 0 {
+		return 0.0
 	}
-	if err := iter.Err(); err != nil && err != io.EOF {
-		return err
-	}
-	return nil
+	num := count*sumSq - math.Pow(sum, 2)
+	return math.Sqrt(num / float64(div))
 }
-
-func (h decodingHandler) Close() {}
--- a/aggregation/counter.go
+++ b/aggregation/counter.go
@@ -20,25 +20,106 @@
 
 package aggregation
 
-import "sync"
+import (
+	"math"
+	"sync"
 
-var (
-	emptyCounter Counter
+	"github.com/m3db/m3metrics/policy"
 )
 
 // Counter aggregates counter values.
 type Counter struct {
-	sum int64 // Sum of the values received
+	Options
+
+	sum   int64
+	sumSq int64
+	count int64
+	max   int64
+	min   int64
 }
 
 // NewCounter creates a new counter.
-func NewCounter() Counter { return emptyCounter }
+func NewCounter(opts Options) Counter {
+	return Counter{
+		Options: opts,
+		max:     math.MinInt64,
+		min:     math.MaxInt64,
+	}
+}
 
-// Add adds a counter value.
-func (c *Counter) Add(value int64) { c.sum += value }
+// Update updates the counter value.
+func (c *Counter) Update(value int64) {
+	c.sum += value
+	if c.IsDefault {
+		return
+	}
+
+	c.count++
+	if c.max < value {
+		c.max = value
+	}
+	if c.min > value {
+		c.min = value
+	}
+
+	if c.IsExpensive {
+		c.sumSq += value * value
+	}
+}
+
+// Count returns the number of values received.
+func (c *Counter) Count() int64 { return c.count }
 
 // Sum returns the sum of counter values.
 func (c *Counter) Sum() int64 { return c.sum }
+
+// SumSq returns the squared sum of counter values.
+func (c *Counter) SumSq() int64 { return c.sumSq }
+
+// Mean returns the mean counter value.
+func (c *Counter) Mean() float64 {
+	if c.count == 0 {
+		return 0
+	}
+	return float64(c.sum) / float64(c.count)
+}
+
+// Stdev returns the standard deviation counter value.
+func (c *Counter) Stdev() float64 {
+	return stdev(float64(c.count), float64(c.sumSq), float64(c.sum))
+}
+
+// Min returns the minimum counter value.
+func (c *Counter) Min() int64 {
+	return c.min
+}
+
+// Max returns the maximum counter value.
+func (c *Counter) Max() int64 {
+	return c.max
+}
+
+// ValueOf returns the value for the aggregation type.
+func (c *Counter) ValueOf(aggType policy.AggregationType) float64 {
+	switch aggType {
+	case policy.Lower:
+		return float64(c.Min())
+	case policy.Upper:
+		return float64(c.Max())
+	case policy.Mean:
+		return c.Mean()
+	case policy.Count:
+		return float64(c.Count())
+	case policy.Sum:
+		return float64(c.Sum())
+	case policy.SumSq:
+		return float64(c.SumSq())
+	case policy.Stdev:
+		return c.Stdev()
+	default:
+		return 0
+	}
+}
 
 // LockedCounter is a locked counter.
 type LockedCounter struct {
@@ -47,7 +128,7 @@ type LockedCounter struct {
 }
 
 // NewLockedCounter creates a new locked counter.
-func NewLockedCounter() *LockedCounter { return &LockedCounter{} }
+func NewLockedCounter(c Counter) *LockedCounter { return &LockedCounter{Counter: c} }
 
 // Reset resets the locked counter.
-func (lc *LockedCounter) Reset() { lc.Counter = emptyCounter }
+func (lc *LockedCounter) Reset(c Counter) { lc.Counter = c }
--- a/aggregation/counter.go
+++ b/aggregation/counter.go
@@ -50,7 +50,7 @@ func NewCounter(opts Options) Counter {
 // Update updates the counter value.
 func (c *Counter) Update(value int64) {
 	c.sum += value
-	if c.IsDefault {
+	if c.UseDefaultAggregation {
 		return
 	}
 
@@ -62,7 +62,7 @@ func (c *Counter) Update(value int64) {
 		c.min = value
 	}
 
-	if c.IsExpensive {
+	if c.HasExpensiveAggregations {
 		c.sumSq += value * value
 	}
 }
@@ -86,18 +86,14 @@ func (c *Counter) Mean() float64 {
 
 // Stdev returns the standard deviation counter value.
 func (c *Counter) Stdev() float64 {
-	return stdev(float64(c.count), float64(c.sumSq), float64(c.sum))
+	return stdev(c.count, float64(c.sumSq), float64(c.sum))
 }
 
 // Min returns the minimum counter value.
-func (c *Counter) Min() int64 {
-	return c.min
-}
+func (c *Counter) Min() int64 { return c.min }
 
 // Max returns the maximum counter value.
-func (c *Counter) Max() int64 {
-	return c.max
-}
+func (c *Counter) Max() int64 { return c.max }
 
 // ValueOf returns the value for the aggregation type.
 func (c *Counter) ValueOf(aggType policy.AggregationType) float64 {
--- a/aggregation/gauge.go
+++ b/aggregation/gauge.go
@@ -20,25 +20,117 @@
 
 package aggregation
 
-import "sync"
+import (
+	"math"
+	"sync"
 
-var (
-	emptyGauge Gauge
+	"github.com/m3db/m3metrics/policy"
+)
+
+const (
+	minFloat64 = -math.MaxFloat64
 )
 
 // Gauge aggregates gauge values.
 type Gauge struct {
-	value float64 // latest value received
+	Options
+
+	last  float64
+	sum   float64
+	sumSq float64
+	count int64
+	max   float64
+	min   float64
 }
 
 // NewGauge creates a new gauge.
-func NewGauge() Gauge { return emptyGauge }
+func NewGauge(opts Options) Gauge {
+	return Gauge{
+		Options: opts,
+		max:     minFloat64,
+		min:     math.MaxFloat64,
+	}
+}
 
-// Set sets the gauge value.
-func (g *Gauge) Set(value float64) { g.value = value }
+// Update updates the gauge value.
+func (g *Gauge) Update(value float64) {
+	g.last = value
+	if g.IsDefault {
+		return
+	}
 
-// Value returns the latest value.
-func (g *Gauge) Value() float64 { return g.value }
+	g.sum += value
+	g.count++
+	if g.max < value {
+		g.max = value
+	}
+	if g.min > value {
+		g.min = value
+	}
+
+	if g.IsExpensive {
+		g.sumSq += value * value
+	}
+}
+
+// Last returns the last value received.
+func (g *Gauge) Last() float64 { return g.last }
+
+// Count returns the number of values received.
+func (g *Gauge) Count() int64 { return g.count }
+
+// Sum returns the sum of gauge values.
+func (g *Gauge) Sum() float64 { return g.sum }
+
+// SumSq returns the squared sum of gauge values.
+func (g *Gauge) SumSq() float64 { return g.sumSq }
+
+// Mean returns the mean gauge value.
+func (g *Gauge) Mean() float64 {
+	if g.count == 0 {
+		return 0.0
+	}
+	return g.sum / float64(g.count)
+}
+
+// Stdev returns the standard deviation gauge value.
+func (g *Gauge) Stdev() float64 {
+	return stdev(float64(g.count), g.sumSq, g.sum)
+}
+
+// Min returns the minimum gauge value.
+func (g *Gauge) Min() float64 {
+	return g.min
+}
+
+// Max returns the maximum gauge value.
+func (g *Gauge) Max() float64 {
+	return g.max
+}
+
+// ValueOf returns the value for the aggregation type.
+func (g *Gauge) ValueOf(aggType policy.AggregationType) float64 {
+	switch aggType {
+	case policy.Last:
+		return g.Last()
+	case policy.Lower:
+		return g.Min()
+	case policy.Upper:
+		return g.Max()
+	case policy.Mean:
+		return g.Mean()
+	case policy.Count:
+		return float64(g.Count())
+	case policy.Sum:
+		return g.Sum()
+	case policy.SumSq:
+		return g.SumSq()
+	case policy.Stdev:
+		return g.Stdev()
+	default:
+		return 0
+	}
+}
 
 // LockedGauge is a locked gauge.
 type LockedGauge struct {
@@ -47,7 +139,7 @@ type LockedGauge struct {
 }
 
 // NewLockedGauge creates a new locked gauge.
-func NewLockedGauge() *LockedGauge { return &LockedGauge{} }
+func NewLockedGauge(g Gauge) *LockedGauge { return &LockedGauge{Gauge: g} }
 
 // Reset resets the locked gauge.
-func (lg *LockedGauge) Reset() { lg.Gauge = emptyGauge }
+func (lg *LockedGauge) Reset(g Gauge) { lg.Gauge = g }
--- a/aggregation/gauge.go
+++ b/aggregation/gauge.go
@@ -55,7 +55,7 @@ func NewGauge(opts Options) Gauge {
 // Update updates the gauge value.
 func (g *Gauge) Update(value float64) {
 	g.last = value
-	if g.IsDefault {
+	if g.UseDefaultAggregation {
 		return
 	}
 
@@ -68,7 +68,7 @@ func (g *Gauge) Update(value float64) {
 		g.min = value
 	}
 
-	if g.IsExpensive {
+	if g.HasExpensiveAggregations {
 		g.sumSq += value * value
 	}
 }
@@ -95,18 +95,14 @@ func (g *Gauge) Mean() float64 {
 
 // Stdev returns the standard deviation gauge value.
 func (g *Gauge) Stdev() float64 {
-	return stdev(float64(g.count), g.sumSq, g.sum)
+	return stdev(g.count, g.sumSq, g.sum)
 }
 
 // Min returns the minimum gauge value.
-func (g *Gauge) Min() float64 {
-	return g.min
-}
+func (g *Gauge) Min() float64 { return g.min }
 
 // Max returns the maximum gauge value.
-func (g *Gauge) Max() float64 {
-	return g.max
-}
+func (g *Gauge) Max() float64 { return g.max }
 
 // ValueOf returns the value for the aggregation type.
 func (g *Gauge) ValueOf(aggType policy.AggregationType) float64 {
--- a/aggregation/options.go
+++ b/aggregation/options.go
@@ -23,16 +23,16 @@ package aggregation
 import "github.com/m3db/m3metrics/policy"
 
 var (
-	defaultOptions = Options{IsDefault: true, IsExpensive: false}
+	defaultOptions = Options{UseDefaultAggregation: true, HasExpensiveAggregations: false}
 )
 
 // Options is the options for aggregations.
 type Options struct {
-	// IsDefault means only default aggregation types are enabled.
-	IsDefault bool
-	// IsExpensive means expensive (multiplication／division)
+	// UseDefaultAggregation means only default aggregation types are enabled.
+	UseDefaultAggregation bool
+	// HasExpensiveAggregations means expensive (multiplication／division)
 	// aggregation types are enabled.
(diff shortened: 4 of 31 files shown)


aggregation/timer.go
```diff
@@ -93,15 +94,53 @@ func (t *Timer) Mean() float64 {
...
 
 // Close closes the timer
 func (t *Timer) Close() {
 	t.stream.Close()
 }
+
+func stdev(count, sumSq, sum float64) float64 {
+	num := count*sumSq - math.Pow(sum, 2)
+	div := count * (count - 1)
+	if div == 0 {
```
prateek: nit: perform the check before calculating num?

aggregation/counter.go
```diff
@@ -20,20 +20,111 @@
...
+}
+
+// Update updates the counter value
+func (c *Counter) Update(value int64) {
+	c.sum += value
+	if !c.customAggregationEnabled {
+		return
+	}
+
+	c.sumSq += math.Pow(float64(value), 2)
```
prateek: Dunno how critical this path is, would be cheaper to bitshift here.

aggregation/counter.go
```diff
@@ -20,20 +20,111 @@
...
+// Mean returns the mean counter value
+func (c *Counter) Mean() int64 {
+	if c.count == 0 {
+		return 0
+	}
+	return c.sum / c.count
+}
+
+// Stdev returns the standard deviation counter value
+func (c *Counter) Stdev() float64 {
```
prateek: nit: rename to `Stddev`

aggregation/gauge.go
```diff
@@ -20,20 +20,126 @@
...
+	min   float64
+
+	customAggregationEnabled bool
 }
 
 // NewGauge creates a new gauge
-func NewGauge() Gauge { return emptyGauge }
+func NewGauge(useCustomeAggregation bool) Gauge {
+	if !useCustomeAggregation {
+		return emptyGauge
```
prateek: Just curious, do you do this to save the cost of allocating a new struct each time?

aggregation/gauge.go
```diff
@@ -20,20 +20,126 @@
...
+// Mean returns the mean gauge value
+func (g *Gauge) Mean() float64 {
+	if g.count == 0 {
+		return 0.0
+	}
+	return g.sum / float64(g.count)
+}
+
+// Stdev returns the standard deviation gauge value
+func (g *Gauge) Stdev() float64 {
```
prateek: nit: rename to `Stddev`

aggregation/timer.go
cw9: will do

aggregation/counter.go
cw9: this is pretty critical, could you elaborate more on the bitshift way?

aggregation/counter.go
cw9: Seems both means the same thing, gonna keep calling it Stdev since it's used in other places in this repo

aggregation/gauge.go
cw9: I think this will return a copy of the empty gauge, which saves the allocation of making a new empty gauge and then copy it out.

aggregation/options.go
```diff
@@ -0,0 +1,56 @@
...
+	return defaultOptions
+}
+
+// ResetSetData resets the aggregation options.
+func (o *Options) ResetSetData(aggTypes policy.AggregationTypes) {
+	o.IsDefault = aggTypes.IsDefault()
+	o.IsExpensive = isExpensive(aggTypes)
+}
+
+func isExpensive(aggTypes policy.AggregationTypes) bool {
```
xichen2020: Move this to `common.go`?

aggregator/options.go
```diff
@@ -122,14 +128,15 @@ func NewOptions() Options {
...
 		timerPrefix:               defaultTimerPrefix,
-		timerSumSuffix:            defaultTimerSumSuffix,
-		timerSumSqSuffix:          defaultTimerSumSqSuffix,
-		timerMeanSuffix:           defaultTimerMeanSuffix,
-		timerLowerSuffix:          defaultTimerLowerSuffix,
-		timerUpperSuffix:          defaultTimerUpperSuffix,
-		timerCountSuffix:          defaultTimerCountSuffix,
-		timerStdevSuffix:          defaultTimerStdevSuffix,
-		timerMedianSuffix:         defaultTimerMedianSuffix,
+		lastSuffix:                defaultAggregationLastSuffix,
```
xichen2020: Right now there is no sum/last suffix for counters/gauges, how do we plan to handle the backward compatibility here?

aggregation/common.go
```diff
@@ -0,0 +1,32 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package aggregation
+
+import "math"
+
+func stdev(count, sumSq, sum float64) float64 {
+	div := count * (count - 1)
```
cw9: good call, will do

aggregation/common.go
```diff
@@ -0,0 +1,32 @@
...
+	if div == 0 {
+		return 0.0
+	}
+	num := count*sumSq - math.Pow(sum, 2)
```
cw9: sure

aggregation/counter.go
```diff
@@ -20,34 +20,115 @@
...
+	"github.com/m3db/m3metrics/policy"
 )
 
 // Counter aggregates counter values.
 type Counter struct {
-	sum int64 // Sum of the values received
+	Options
+
+	sum   int64
+	sumSq int64
```
cw9: sounds good

aggregator/options.go
cw9: The default aggregation will not add suffix in elem.go for backward compatibility

aggregation/options.go
```diff
@@ -0,0 +1,56 @@
...
+package aggregation
+
+import "github.com/m3db/m3metrics/policy"
+
+var (
+	defaultOptions = Options{IsDefault: true, IsExpensive: false}
+)
+
+// Options is the options for aggregations.
+type Options struct {
```
cw9: Yes it's used in elem.go as well

aggregator/elem.go
```diff
@@ -392,9 +458,13 @@ func (e *TimerElem) Close() {
 	e.values = e.values[:0]
 	e.toConsume = e.toConsume[:0]
 	pool := e.opts.TimerElemPool()
+	aggTypesPool := e.opts.AggregationTypesPool()
+	quantileFloatsPool := e.opts.QuantileFloatsPool()
 	e.Unlock()
 
 	pool.Put(e)
+	aggTypesPool.Put(e.aggTypes)
```
xichen2020: Same here, the internal fields need to be returned to pool before the element is returned.

aggregator/elem.go
```diff
@@ -392,9 +458,13 @@ func (e *TimerElem) Close() {
...
+	quantileFloatsPool.Put(e.quantiles)
```
xichen2020: Need to distinguish pooled and non-pooled quantiles.

aggregator/elem.go
```diff
@@ -469,52 +539,53 @@ func (e *TimerElem) indexOfWithLock(alignedStart int64) (int, bool) {
...
-	fn(fullTimerPrefix, e.id, timerLowerSuffix, timeNanos, agg.Min(), e.policy)
-	fn(fullTimerPrefix, e.id, timerUpperSuffix, timeNanos, agg.Max(), e.policy)
-	fn(fullTimerPrefix, e.id, timerCountSuffix, timeNanos, float64(agg.Count()), e.policy)
-	fn(fullTimerPrefix, e.id, timerStdevSuffix, timeNanos, agg.Stdev(), e.policy)
-	for idx, q := range quantiles {
-		v := agg.Quantile(q)
-		if q == 0.5 {
-			fn(fullTimerPrefix, e.id, timerMedianSuffix, timeNanos, v, e.policy)
+	fullTimerPrefix := e.opts.FullTimerPrefix()
+	if e.aggOpts.UseDefaultAggregation {
```
xichen2020: Would be less nesting if you handle `if !e.aggOpts.UseDefaultAggregation` here.

aggregator/elem.go
```diff
@@ -569,9 +640,11 @@ func (e *GaugeElem) Close() {
 	e.values = e.values[:0]
 	e.toConsume = e.toConsume[:0]
 	pool := e.opts.GaugeElemPool()
+	aggTypesPool := e.opts.AggregationTypesPool()
 	e.Unlock()
 
 	pool.Put(e)
+	aggTypesPool.Put(e.aggTypes)
```
xichen2020: Same here.

aggregator/options.go
```diff
@@ -146,7 +153,7 @@ func NewOptions() Options {
 		entryCheckInterval:        defaultEntryCheckInterval,
 		entryCheckBatchPercent:    defaultEntryCheckBatchPercent,
 		maxTimerBatchSizePerWrite: defaultMaxTimerBatchSizePerWrite,
-		defaultPolicies:           defaultDefaultPolicies,
+		defaultPolicies:           defaultPolicies,
```
xichen2020: hm `defaultDefaultPolicies` means the default value for the default policies, should we change it back?

aggregator/options.go
```diff
@@ -526,6 +543,26 @@ func (o *options) BufferedEncoderPool() msgpack.BufferedEncoderPool {
...
+	opts := *o
+	opts.aggTypesPool = pool
+	return &opts
+}
+
+func (o *options) AggregationTypesPool() policy.AggregationTypesPool {
+	return o.aggTypesPool
+}
+
+func (o *options) SetQuantileFloatsPool(pool pool.FloatsPool) Options {
```
xichen2020: Could jsut be `QuantilesPool`

aggregator/options.go
```diff
@@ -612,7 +656,9 @@ func (o *options) computeFullGaugePrefix() {
 
 // By default we use e.g. ".p50", ".p95", ".p99" for the 50th/95th/99th percentile.
 func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	return []byte(fmt.Sprintf(".p%0.0f", quantile*100))
+	s := fmt.Sprintf(".p%0.0f", quantile*10000)
```
xichen2020: This makes assumptions about what quantiles we'll support, can we instead write a function to keep multiplying the quantile by 10 until it becomes an integer and use that in the string?

aggregator/types.go
```diff
@@ -316,30 +323,42 @@ type Options interface {
...
 	BufferedEncoderPool() msgpack.BufferedEncoderPool
 
+	// SetAggregationTypesPool sets the aggregation types pool.
+	SetAggregationTypesPool(pool policy.AggregationTypesPool) Options
+
+	// AggregationTypesPool returns the aggregation types pool.
+	AggregationTypesPool() policy.AggregationTypesPool
+
+	// SetQuantileFloatsPool sets the timer quantiles pool.
+	SetQuantileFloatsPool(pool pool.FloatsPool) Options
```
xichen2020: Could just be `SetQuantilesPool`?

aggregator/options.go
cw9: It's not easy to find the integer value for a float, see: https://play.golang.org/p/SKBCfflTwD, also if we want to keep multiplying until we reach a "integer" we would need to assume a magic epsilon to check if a float is integer or not anyway, so I'm just gonna assume the precision instead. Will change the trimming to be more dynamic.

glide.lock
```diff
@@ -151,6 +151,8 @@ imports:
   - m3/customtransports
   - m3/thrift
   - m3/thriftudp
+- name: github.com/willf/bitset
```
cw9: sure

aggregator/elem.go
```diff
@@ -158,23 +170,52 @@ func (e *elemBase) MarkAsTombstoned() {
...
+		return e.opts.AggregationUpperSuffix()
+	case policy.Mean:
+		return e.opts.AggregationMeanSuffix()
+	case policy.Median:
+		return e.opts.AggregationMedianSuffix()
+	case policy.Count:
+		return e.opts.AggregationCountSuffix()
+	case policy.Sum:
+		return e.opts.AggregationSumSuffix()
+	case policy.SumSq:
```
cw9: sure

aggregator/elem.go
```diff
@@ -230,9 +271,11 @@ func (e *CounterElem) Close() {
 	e.values = e.values[:0]
 	e.toConsume = e.toConsume[:0]
 	pool := e.opts.CounterElemPool()
+	aggTypesPool := e.opts.AggregationTypesPool()
 	e.Unlock()
 
 	pool.Put(e)
+	aggTypesPool.Put(e.aggTypes)
```
cw9: good catch, will do

aggregator/elem.go
```diff
@@ -306,20 +349,43 @@ func (e *CounterElem) indexOfWithLock(alignedStart int64) (int, bool) {
...
+func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
+	e.elemBase.ResetSetData(id, sp, aggTypes)
+	e.quantiles = e.quantilesFromAggTypes(aggTypes)
+}
+
+func (e *TimerElem) quantilesFromAggTypes(aggTypes policy.AggregationTypes) []float64 {
+	if aggTypes.IsDefault() {
+		return e.opts.TimerQuantiles()
+	}
+	return aggTypes.PooledQuantiles(e.opts.QuantileFloatsPool())
```
cw9: will do

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sure

aggregator/types.go
cw9: sure

cw9:
I disabled default timer quantiles override with config and allowed aggregation types override for timer with config because if a user sets some random timer quantiles as default we will have backward incompatible issues that they could no longer get those quantiles once they started to use custom aggregation types.

aggregator/elem.go
```diff
@@ -360,6 +335,15 @@ func (e *CounterElem) processValue(timeNanos int64, agg aggregation.Counter, fn
 	}
 }
 
+// NB(cw) suffix overrides default suffixes to not add
+// suffix for default Counter metrics for backward compatibility reasons.
+func (e *CounterElem) suffix(aggType policy.AggregationType) []byte {
+	if aggType == policy.Sum {
```
xichen2020: Should we add `defaultCounterAggregationType` to the options and compare against it instead of a hardcoded policy here?

aggregator/elem.go
```diff
@@ -728,3 +691,12 @@ func (e *GaugeElem) processValue(timeNanos int64, agg aggregation.Gauge, fn aggM
 		fn(e.opts.FullGaugePrefix(), e.id, e.suffix(aggType), timeNanos, agg.ValueOf(aggType), e.sp)
 	}
 }
+
+// NB(cw) suffix overrides default suffixes to not add
+// suffix for default Gauge metrics for backward compatibility reasons.
+func (e *GaugeElem) suffix(aggType policy.AggregationType) []byte {
+	if aggType == policy.Last {
```
xichen2020: Should we add `defaultGaugeAggregationType` to the options and compare against it?

aggregator/options.go
```diff
@@ -65,13 +60,26 @@ var (
...
 	defaultMaxTimerBatchSizePerWrite = 0
-	defaultPolicies                  = []policy.Policy{
+	defaultDefaultPolicies           = []policy.Policy{
 		policy.NewPolicy(policy.NewStoragePolicy(10*time.Second, xtime.Second, 2*24*time.Hour), policy.DefaultAggregationID),
 		policy.NewPolicy(policy.NewStoragePolicy(time.Minute, xtime.Minute, 40*24*time.Hour), policy.DefaultAggregationID),
 	}
-	defaultTimerQuantiles = []float64{0.5, 0.95, 0.99}
 
-	errInvalidQuantiles = fmt.Errorf("quantiles must be nonempty and between %f and %f", minQuantile, maxQuantile)
+	defaultTimerAggregationTypes = policy.AggregationTypes{
```
xichen2020: `defaultDefaultTimerAggregationTypes` might be better?

aggregator/options.go
```diff
@@ -88,6 +96,7 @@ type options struct {
 	countSuffix               []byte
 	stdevSuffix               []byte
 	medianSuffix              []byte
+	timerAggTypes             policy.AggregationTypes
```
xichen2020: `defaultTimerAggregationTypes` might be better?

aggregator/options.go
```diff
@@ -575,8 +572,12 @@ func (o *options) FullGaugePrefix() []byte {
 	return o.fullGaugePrefix
 }
 
-func (o *options) TimerQuantileSuffixes() [][]byte {
-	return o.timerQuantileSuffixes
+func (o *options) TimerQuantiles() []float64 {
+	return o.timerQuantiles
+}
+
+func (o *options) Suffix(aggtype policy.AggregationType) []byte {
```
xichen2020: `aggtype` -> `aggType`?

xichen2020:
@cw9 the default timer quantiles is a global setting that applies to all timers and cannot be changed by end users (just like what statsite has). If the users opt in to use custom aggregation functions, it means they have complete control of what aggregation types they choose for their timers and therefore the default quantiles are no longer applicable to their timers. 

In addition, looking up the aggregation type multiple timers for every timer flush will impose non-trivial performance overhead on the agg tier due to map lookup overhead. I think we should leave it as it to avoid map lookups for performance reasons.

aggregator/options.go
```diff
@@ -65,13 +60,26 @@ var (
...
+		policy.Upper,
+		policy.Count,
+		policy.Stdev,
+		policy.Median,
+		policy.P50,
+		policy.P95,
+		policy.P99,
+	}
+
+	defaultQuantilePrecision = 0.0000000001 // 9 zeros after 0.
```
xichen2020: nit: `1e-9`?

aggregator/options.go
```diff
@@ -116,10 +125,10 @@ type options struct {
...
 
 	// Derived options.
-	fullCounterPrefix     []byte
-	fullTimerPrefix       []byte
-	fullGaugePrefix       []byte
-	timerQuantileSuffixes [][]byte
+	fullCounterPrefix []byte
+	fullTimerPrefix   []byte
+	fullGaugePrefix   []byte
+	suffixMap         map[policy.AggregationType][]byte
```
xichen2020: Map lookups for suffixes are expensive, I'm concerned about doing map lookups multiple times for each timer aggregation given that the timer flushes is the biggest performance bottlenecks on the agg tier.

aggregator/options.go
```diff
@@ -656,9 +686,15 @@ func (o *options) computeFullGaugePrefix() {
...
 func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	s := fmt.Sprintf(".p%0.0f", quantile*10000)
-	// Trim 2 rightmost "0"s.
-	return []byte(strings.TrimSuffix(strings.TrimSuffix(s, "0"), "0"))
+	s := fmt.Sprintf("%0.0f", quantile/defaultQuantilePrecision)
+	return []byte(fmt.Sprintf(".p%s", trimQuantileSuffix(s)))
+}
+
+func trimQuantileSuffix(s string) string {
+	for strings.HasSuffix(s, "0") && len(s) > 2 {
```
xichen2020: hm I don't think this works for say, the 5th percentile.

Perhaps use something like https://play.golang.org/p/VXrfQm8A9p?

xichen2020:
Also can we write an integration test that contains aggregation types so we know it's working as expected?

aggregator/elem.go
cw9: I don't think it should be set in the aggregator options, because the assumption was also made in the aggregation package

aggregator/elem.go
cw9: same

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sounds good, will provide a way to access the slice of default suffixes

aggregator/options.go
cw9: Good catch, will fix

aggregator/elem.go
```diff
@@ -360,6 +335,15 @@ func (e *CounterElem) processValue(timeNanos int64, agg aggregation.Counter, fn
 	}
 }
 
+// NB(cw) suffix overrides default suffixes to not add
```
cw9: Created: https://github.com/m3db/m3aggregator/issues/26

aggregator/elem.go
```diff
@@ -306,20 +325,56 @@ func (e *CounterElem) indexOfWithLock(alignedStart int64) (int, bool) {
 }
 
 func (e *CounterElem) processValue(timeNanos int64, agg aggregation.Counter, fn aggMetricFn) {
-	fn(e.opts.FullCounterPrefix(), e.id, nil, timeNanos, float64(agg.Sum()), e.policy)
+	var fullCounterPrefix = e.opts.FullCounterPrefix()
+	if e.aggOpts.UseDefaultAggregation {
+		// No suffix for default aggregations.
+		fn(fullCounterPrefix, e.id, e.suffix(policy.Sum), timeNanos, float64(agg.Sum()), e.sp)
```
xichen2020: Put a TODO here to make default agg types for counters configurable.

aggregator/elem.go
```diff
@@ -645,5 +699,23 @@ func (e *GaugeElem) indexOfWithLock(alignedStart int64) (int, bool) {
 }
 
 func (e *GaugeElem) processValue(timeNanos int64, agg aggregation.Gauge, fn aggMetricFn) {
-	fn(e.opts.FullGaugePrefix(), e.id, nil, timeNanos, agg.Value(), e.policy)
+	var fullGaugePrefix = e.opts.FullGaugePrefix()
+	if e.aggOpts.UseDefaultAggregation {
+		// No suffix for default aggregations.
+		fn(fullGaugePrefix, e.id, e.suffix(policy.Last), timeNanos, agg.Last(), e.sp)
```
xichen2020: Put a TODO here to make default agg types for gauges configurable.

aggregation/timer.go
```diff
@@ -21,85 +21,113 @@
...
 
-// Add adds a timer value
+// Add adds a timer value.
 func (t *Timer) Add(value float64) {
 	t.count++
 	t.sum += value
-	t.sumSq += value * value
 	t.stream.Add(value)
+
+	if t.UseDefaultAggregation || t.HasExpensiveAggregations {
```
xichen2020: Can probably use a single flag?

aggregator/elem.go
```diff
@@ -306,20 +325,56 @@ func (e *CounterElem) indexOfWithLock(alignedStart int64) (int, bool) {
...
 		values:   make([]timedTimer, 0, defaultNumValues), // in most cases values will have two entries
 	}
+	e.ResetSetData(id, sp, aggTypes)
+	return e
+}
+
+// ResetSetData overrides elemBase.ResetData, to include quantile update for timer elements.
+func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
+	if aggTypes.IsDefault() {
+		aggTypes = e.opts.DefaultTimerAggregationTypes()
```
xichen2020: If we end up making the default agg types configurable for counters and gauges, could make `isAggTypesPooled` to `elemBase` and the logic that determines whether the agg type is pooled into the caller.

aggregator/options.go
xichen2020: As discussed offline, could preallocated a `[][]byte` up to the max id and have an `RawID` function that indexes into the `[][]byte` to get the raw id back.

aggregator/options.go
```diff
@@ -612,7 +698,15 @@ func (o *options) computeFullGaugePrefix() {
 
 // By default we use e.g. ".p50", ".p95", ".p99" for the 50th/95th/99th percentile.
 func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	return []byte(fmt.Sprintf(".p%0.0f", quantile*100))
+	return []byte(".p" + addSuffixZero(strconv.FormatFloat(quantile, 'f', -1, 64)[2:]))
```
xichen2020: Add some comments here explaining what is being done? Also might wanna get rid of the leading zeros after the decimal point (e.g., `p5` is probably better than `p05`)

services/m3aggregator/config/aggregator.go
```diff
@@ -155,6 +157,12 @@ type AggregatorConfiguration struct {
 
 	// Pool of buffered encoders.
 	BufferedEncoderPool pool.ObjectPoolConfiguration `yaml:"bufferedEncoderPool"`
+
+	// Pool of aggregation types.
+	AggregationTypesPool pool.ObjectPoolConfiguration `yaml:"aggregationTypesPool"`
+
+	// Pool of quantile slices.
+	QuantilesFloatsPool pool.BucketizedPoolConfiguration `yaml:"quantilesFloatsPool"`
```
xichen2020: Just `QuantilesPool`?

services/m3aggregator/config/aggregator.go
```diff
@@ -307,22 +322,36 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
...
+	iOpts = instrumentOpts.SetMetricsScope(scope.SubScope("aggregation-types-pool"))
+	aggTypesPoolOpts := c.AggregationTypesPool.NewObjectPoolOptions(iOpts)
+	aggTypesPool := policy.NewAggregationTypesPool(aggTypesPoolOpts)
+	opts = opts.SetAggregationTypesPool(aggTypesPool)
+	aggTypesPool.Init(func() policy.AggregationTypes {
+		return make(policy.AggregationTypes, 0, len(policy.ValidAggregationTypes))
+	})
+
+	// Set quantile floats pool.
+	iOpts = instrumentOpts.SetMetricsScope(scope.SubScope("quantile-floats-pool"))
```
xichen2020: Just `quantiles-pool`?

aggregator/elem.go
```diff
@@ -130,10 +136,19 @@ type GaugeElem struct {
...
+func newElemBase(opts Options) elemBase {
+	return elemBase{
+		opts:    opts,
+		aggOpts: aggregation.NewOptions(),
+	}
+}
+
 // ResetSetData resets the counter and sets data.
-func (e *elemBase) ResetSetData(id id.RawID, policy policy.Policy) {
+func (e *elemBase) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
```
xichen2020: Would it be easier to pass in `policy.Policy` instead storage policy and aggregation types separately?

services/m3aggregator/config/aggregator.go
cw9: Sure

services/m3aggregator/config/aggregator.go
cw9: Sure

aggregator/options.go
cw9: I would prefer to do p05 than p5, we need to differentiate quantiles like p05 and p0005. Will add a comment

cw9: talked offline, gonna do the format of p0.05 and p99.99

aggregation/timer.go
cw9: Sure

aggregator/elem.go
cw9: We can't directly use policy.Policy here, that would require decompress the policy.AggregationID to get the list of aggregation types

aggregator/elem.go
cw9: Sounds good

aggregator/options.go
```diff
@@ -126,7 +126,7 @@ type options struct {
 	fullCounterPrefix []byte
 	fullTimerPrefix   []byte
 	fullGaugePrefix   []byte
-	suffixMap         map[policy.AggregationType][]byte
+	suffixSlice       [][]byte
```
xichen2020: Just `suffixes`?

aggregator/elem.go
xichen2020: Right I see

aggregation/options.go
```diff
@@ -0,0 +1,47 @@
...
+	HasExpensiveAggregations bool
+}
+
+// NewOptions creates a new aggregation options.
+func NewOptions() Options {
+	return defaultOptions
+}
+
+// ResetSetData resets the aggregation options.
+func (o *Options) ResetSetData(aggTypes policy.AggregationTypes) {
```
xichen2020: This is only called when element is created / fetched from pool yeah?

aggregator/options.go
cw9: Sure

aggregation/options.go
cw9: Yes, which is intended, if the default/custom aggregation types for this timer does not want stdev or sumsq, HasExpensiveOperations will be reset to false and skip the sumsq calculation.

PR #17, opened 2017-05-22
Shard metrics within aggregation server

xichen2020:
cc @robskillington @cw9 @jeromefroe @prateek 

Currently all the metrics are stored in a global map, which in turn also stores the element for each metric in a single list per resolution. While this works for small to medium qps, it doesn't scale as the number of unique metrics grows because we have SLAs on the list flush latency for each resolution, which worsens as there are more and more metrics put on the list.

This PR adds logic to shard metrics within each aggregation server so that each shard has its own map / list to store aggregated metrics. This reduce lock contention among metrics on the map, as well as the number of elements per list, which helps with the flush latency.

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -22,10 +22,14 @@ package aggregator
 
 import (
 	"errors"
+	"fmt"
+	"math"
 	"sync"
-	"sync/atomic"
 	"time"
 
+	"github.com/m3db/m3cluster/services"
+	"github.com/m3db/m3cluster/services/placement"
+	"github.com/m3db/m3metrics/metric/id"
 	"github.com/m3db/m3metrics/metric/unaggregated"
 	"github.com/m3db/m3metrics/policy"
 	"github.com/m3db/m3x/clock"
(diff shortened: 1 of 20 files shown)


services/m3aggregator/config/aggregator.go
```diff
@@ -160,7 +180,40 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
...
-	// Set flushing options.
+	// Set instance id.
+	instanceID, err := instanceID(address)
+	if err != nil {
+		return nil, err
+	}
+	opts = opts.SetInstanceID(instanceID)
+
+	// Create kv client.
+	iOpts = instrumentOpts.SetMetricsScope(instrumentOpts.MetricsScope().SubScope("kvClient"))
```
cw9: nit: 
`iOpts = instrumentOpts.SetMetricsScope(scope.SubScope("kvClient"))`

services/m3aggregator/config/aggregator.go
```diff
@@ -385,3 +529,15 @@ func setMetricPrefixOrSuffix(
 	}
 	return fn([]byte(str))
 }
+
+func instanceID(address string) (string, error) {
```
cw9: why not just use the address? what's in the first part of address?

aggregator/aggregator.go
```diff
@@ -34,116 +38,362 @@ import (
...
 		agg.metrics.addMetricWithPoliciesList.ReportError(agg.nowFn().Sub(callStart))
-		return errAggregatorClosed
+		return err
+	}
+	shard, err := agg.shardFor(mu.ID)
+	if err != nil {
+		agg.metrics.addMetricWithPoliciesList.ReportError(agg.nowFn().Sub(callStart))
+		return err
+	}
+	err = shard.AddMetricWithPoliciesList(mu, pl)
```
xichen2020: Not sure I fully understand the scenario, when a placement change happens after the shard is determined:
* if the shard is still owned by the aggregator, it'll continue to write to that shard.
* if the shard is no longer owned by the aggregator or the number of shards changed, the shard will close itself asynchronously as a result, and the shard.Close function will make sure all inflight writes to that shard are completed before closing itself.

aggregator/aggregator.go
```diff
@@ -160,34 +410,23 @@ func (agg *aggregator) tick() {
...
+		tickResult           tickResult
+	)
+	for _, shard := range ownedShards {
+		shardTickResult := shard.Tick(perShardTickDuration)
+		tickResult = tickResult.merge(shardTickResult)
+	}
 	tickDuration := agg.nowFn().Sub(start)
-	agg.metrics.tickExpired.Inc(tickExpired)
-	agg.metrics.tickDuration.Record(tickDuration)
+	agg.metrics.tick.Report(tickResult, tickDuration)
```
xichen2020: Sure thing, will do

services/m3aggregator/config/aggregator.go
xichen2020: Sounds good.

services/m3aggregator/config/aggregator.go
xichen2020: Currently the address is "0.0.0.0:1234" or "127.0.0.1:1234", there is no hostname in it.

aggregator/list.go
```diff
@@ -63,6 +61,7 @@ func newMetricListMetrics(scope tally.Scope) metricListMetrics {
 }
 
 type encodeFn func(mp aggregated.ChunkedMetricWithPolicy) error
+type waitForFn func(d time.Duration) <-chan time.Time
```
cw9: is this still used?

aggregator/types.go
```diff
@@ -198,6 +205,12 @@ type Options interface {
 	// GaugePrefix returns the prefix for gauges
 	GaugePrefix() []byte
 
+	// SetTimeLock sets the time lock
+	SetTimeLock(value *sync.RWMutex) Options
```
cw9: Add period

xichen2020: Will put up another diff to add periods to all comments.

PR #18, opened 2017-05-26
Coordinate flushes across aggregated metric lists

xichen2020:
cc @robskillington @cw9 @jeromefroe @prateek 

This PR adds logic to coordinate flushes across aggregated metric lists in different shards in order to reduce spikiness in p99 flush latencies by controlling the concurrency of flushes and adding jitters to different flushes.

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -148,7 +148,8 @@ type aggregator struct {
 	instanceID    string
 	shardFn       ShardFn
 	checkInterval time.Duration
-	handler       Handler
+	flushManager  FlushManager
+	flushHandler  Handler
 
 	shardIDs              []uint32
 	shards                []*aggregatorShard
@@ -174,7 +175,8 @@ func NewAggregator(opts Options) Aggregator {
 		instanceID:            opts.InstanceID(),
 		shardFn:               opts.ShardFn(),
 		checkInterval:         opts.EntryCheckInterval(),
-		handler:               opts.FlushHandler(),
+		flushManager:          opts.FlushManager(),
+		flushHandler:          opts.FlushHandler(),
 		placementWatcher:      placementWatcher,
 		placementCutoverNanos: uninitializedCutoverNanos,
 		metrics:               metrics,
@@ -235,7 +237,8 @@ func (agg *aggregator) Close() error {
 	for _, shardID := range agg.shardIDs {
 		agg.shards[shardID].Close()
 	}
-	agg.handler.Close()
+	agg.flushManager.Close()
+	agg.flushHandler.Close()
 	return nil
 }
 
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -93,6 +93,8 @@ func (e *Entry) IncWriter() { atomic.AddInt32(&e.numWriters, 1) }
 func (e *Entry) DecWriter() { atomic.AddInt32(&e.numWriters, -1) }
 
 // ResetSetData resets the entry and sets initial data.
+// NB(xichen): we need to reset the options here to use the correct
+// time lock contained in the options.
 func (e *Entry) ResetSetData(lists *metricLists, opts Options) {
 	e.closed = false
 	e.opts = opts
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -62,6 +62,7 @@ func newEntryMetrics(scope tally.Scope) entryMetrics {
 type Entry struct {
 	sync.RWMutex
 
+	closed                 bool
 	opts                   Options
 	hasDefaultPoliciesList bool
 	useDefaultPolicies     bool
@@ -71,7 +72,6 @@ type Entry struct {
 	numWriters             int32
 	lastAccessNanos        int64
 	aggregations           map[policy.Policy]*list.Element
-	closed                 bool
 	metrics                entryMetrics
 }
 
@@ -79,11 +79,10 @@ type Entry struct {
 func NewEntry(lists *metricLists, opts Options) *Entry {
 	scope := opts.InstrumentOptions().MetricsScope().SubScope("entry")
 	e := &Entry{
-		opts:         opts,
 		aggregations: make(map[policy.Policy]*list.Element),
 		metrics:      newEntryMetrics(scope),
 	}
-	e.ResetSetData(lists)
+	e.ResetSetData(lists, opts)
 	return e
 }
 
@@ -94,8 +93,9 @@ func (e *Entry) IncWriter() { atomic.AddInt32(&e.numWriters, 1) }
 func (e *Entry) DecWriter() { atomic.AddInt32(&e.numWriters, -1) }
 
 // ResetSetData resets the entry and sets initial data.
-func (e *Entry) ResetSetData(lists *metricLists) {
+func (e *Entry) ResetSetData(lists *metricLists, opts Options) {
 	e.closed = false
+	e.opts = opts
 	e.hasDefaultPoliciesList = false
 	e.useDefaultPolicies = false
 	e.cutoverNanos = uninitializedCutoverNanos
--- a/aggregator/flush_mgr.go
+++ b/aggregator/flush_mgr.go
@@ -1,0 +1,338 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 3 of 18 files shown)


aggregator/aggregator.go
```diff
@@ -148,7 +148,8 @@ type aggregator struct {
 	instanceID    string
 	shardFn       ShardFn
 	checkInterval time.Duration
-	handler       Handler
+	flushManager  FlushManager
```
cw9: it's a bit weird that aggregator keeps ref to flushMgr and flushHandler just to close them, maybe they should be closed in main() or serve()?

aggregator/entry.go
```diff
@@ -94,8 +93,9 @@ func (e *Entry) IncWriter() { atomic.AddInt32(&e.numWriters, 1) }
 func (e *Entry) DecWriter() { atomic.AddInt32(&e.numWriters, -1) }
 
 // ResetSetData resets the entry and sets initial data.
-func (e *Entry) ResetSetData(lists *metricLists) {
+func (e *Entry) ResetSetData(lists *metricLists, opts Options) {
```
cw9: add comment that entry needs to reset opts because of the per shard timelock?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+
+type flushManager struct {
+	sync.Mutex
+
+	nowFn         clock.NowFn
+	checkEvery    time.Duration
+	jitterEnabled bool
+	workers       xsync.WorkerPool
+	scope         tally.Scope
+
```
cw9: do we need the empty line here?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+	return mgr
+}
+
+func (mgr *flushManager) Register(flusher PeriodicFlusher) error {
+	mgr.Lock()
+	bucket, err := mgr.bucketForWithLock(flusher)
+	if err == nil {
+		bucket.Add(flusher)
+		mgr.Unlock()
+		return nil
```
cw9: save the unlock and return nil, just use return err outside?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+		return nil, errFlushManagerClosed
+	}
+	flushInterval := l.FlushInterval()
+	for _, bucket := range mgr.buckets {
+		if bucket.interval == flushInterval {
+			return bucket, nil
+		}
+	}
+	bucketScope := mgr.scope.SubScope("bucket").Tagged(map[string]string{
+		"interval": flushInterval.String(),
```
cw9: mgr.scope.Tagged("bucket_interval", flushInterval.String())?
or 
mgr.scope.Tagged("flush_bucket", flushInterval.String())?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+
+// NB(xichen): apparently timer.Reset() is more difficult to use than I originally
+// anticipated. For now I'm simply waking up every second to check for updates. Maybe
+// when I have more time I'll spend a few hours to get timer.Reset() right and switch
+// to a timer.Start + timer.Stop + timer.Reset + timer.After based approach.
+func (mgr *flushManager) flush() {
+	defer mgr.wgFlush.Done()
+
+	for {
+		var (
```
cw9: define these before for loop?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+	// heap for the flush goroutine to pick it up.
+	nowNanos := mgr.nowNanos()
+	nextFlushNanos := nowNanos + int64(flushInterval)
+	if mgr.jitterEnabled {
+		jitterNanos := int64(mgr.rand.Int63n(int64(flushInterval)))
+		nextFlushNanos = nowNanos + jitterNanos
+	}
+	newFlushMetadata := flushMetadata{
+		timeNanos: nextFlushNanos,
+		bucketIdx: len(mgr.buckets) - 1,
```
cw9: why not use a map for buckets keyed on interval? so here can reference to the interval?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+// flushMetadata contains metadata information for a flush.
+type flushMetadata struct {
+	timeNanos int64
+	bucketIdx int
+}
+
+// flushMetadataHeap is a min heap for flush metadata where the metadata with the
+// earliest flush time will be at the top of the heap. Unlike the generic heap in
+// the container/heap package, pushing data to or popping data off of the heap doesn't
+// require conversion between flush metadata and interface{}, therefore avoiding the
```
cw9: do we need the custom impl here? the heap is not used that often and the length isn't too big?

aggregator/flush_mgr.go
```diff
@@ -0,0 +1,338 @@
...
+		go mgr.flush()
+	}
+	return mgr
+}
+
+func (mgr *flushManager) Register(flusher PeriodicFlusher) error {
+	mgr.Lock()
+	bucket, err := mgr.bucketForWithLock(flusher)
+	if err == nil {
+		bucket.Add(flusher)
```
cw9: this means if the flusher has a one min interval and it registered it self 10 seconds after the last one min bucket was flushed, this flusher maybe flushed in 50 second later? is that ok?

aggregator/flush_mgr.go
xichen2020: I usually separate the internal fields from the arguments/options so I can keep track which are consts and which aren't

aggregator/flush_mgr.go
xichen2020: Yeah that's ok, each element in the list themselves have a map that are keyed on the actual timestamps of the aggregated metrics, so it's safe to flush the list at any time.

aggregator/flush_mgr.go
xichen2020: meh I'd rather be explicit about the fact this is a no error case.

aggregator/flush_mgr.go
xichen2020: `bucket` is already in the name of the metric, I can change the tag name to `flush-interval` if that's better?

aggregator/flush_mgr.go
xichen2020: Map lookups are much slower than indexing into slices. Also GC overhead on maps is bigger than slices of the same size due to more complex internal implementation so I'd avoid maps if I could use a slice.

aggregator/flush_mgr.go
xichen2020: I could, but it feels more clear this way though, the compiler should take care of resetting it per iteration.

aggregator/flush_mgr.go
xichen2020: The heap is used per flush, which could be once a second for metrics with 1s resolution. Also given the not-so-complex heap implementation I'd rather reduce GC as much as possible.

aggregator/aggregator.go
xichen2020: I don't think it's too big an issue, since the aggregator is the owner of these. If golang has destructors and unique_ptr objects, this is like passing in an unique_ptr into the aggregator and the aggregator destructor destroys both when they go out of scope. Unfortunately we don't have destructors or unique_ptr objects, so we have to do it inside `Close`.

aggregator/entry.go
xichen2020: Sure thing.

aggregator/flush_mgr.go
cw9: I meant to remove .SubScope("bucket") call and only used the Tagged()

aggregator/flush_mgr.go
cw9: sg

aggregator/flush_mgr.go
cw9: sg

aggregator/flush_mgr.go
cw9: sg

aggregator/aggregator.go
cw9: sg

aggregator/flush_mgr.go
cw9: I'm not feeling strong about either way, but it feels clearer to define a variable outside the loop to indicate it's reused in the loop.

PR #19, opened 2017-05-30
Various performance optimizations and bug fixes

xichen2020:
cc @robskillington @cw9 @prateek @jeromefroe 

This PR contains various performance optimizations for the aggregation tier. Specifically, it includes
* Adding an option to provide a nil sample pool, which effectively disables sample pooling. The sample pool is used to provide sample objects for every timer value, which is on the critical path and is heavily contended. Profile shows getting values from the pool and putting values into the pool accounts for >30% of the CPU due to heavy contention. Disabling sample pooling significantly reduced write p99 (4ms -> 1ms) and flush p99 (1.5s -> 200ms).
* Adding an option to periodically flush streams at write time, therefore reducing processing load during global flushing. This has significantly improved flush p99 for aggregation servers that have large batches of timer values. In one case, we saw 5s -> 150ms reduction w.r.t. p99 flush latency.
* Replace math.Pow(x, 2) with x * x. This used to account for 4% of the CPU due to math.Pow(x, 2) being 100x more expensive than the plain multiplication. 
* Preallocate enough buffer for the buffered encoder. The buffered encode is used to hold the aggregated metrics. Previous it starts with an empty buffer and gradually grows to the maximum flush size, which requires unnecessary copy and reallocations.

It also includes more instrumentation and bug fixes. Specifically, it includes:
* Fixing a memory leak in timer creation where the underlying stream is not reset, causing the stream (and the float slices and the samples) to not be properly returned to pool.
* Adding logging to record the remote address when a decode error happens. This has allowed us to determine the primary cause of the decode errors is due to misconfigured hosts in the wrong datacenter.

### DIFF
--- a/aggregation/quantile/cm/options.go
+++ b/aggregation/quantile/cm/options.go
@@ -28,12 +28,13 @@ import (
 )
 
 const (
-	minEps          = 0.0
-	maxEps          = 0.5
-	minQuantile     = 0.0
-	maxQuantile     = 1.0
-	defaultEps      = 1e-3
-	defaultCapacity = 16
+	minEps            = 0.0
+	maxEps            = 0.5
+	minQuantile       = 0.0
+	maxQuantile       = 1.0
+	defaultEps        = 1e-3
+	defaultCapacity   = 16
+	defaultFlushEvery = 0
 )
 
 var (
@@ -52,6 +53,7 @@ type options struct {
 	eps        float64
 	quantiles  []float64
 	capacity   int
+	flushEvery int
 	streamPool StreamPool
 	samplePool SamplePool
 	floatsPool pool.FloatsPool
@@ -60,9 +62,10 @@ type options struct {
 // NewOptions creates a new options
 func NewOptions() Options {
 	o := &options{
-		eps:       defaultEps,
-		quantiles: defaultQuantiles,
-		capacity:  defaultCapacity,
+		eps:        defaultEps,
+		quantiles:  defaultQuantiles,
+		capacity:   defaultCapacity,
+		flushEvery: defaultFlushEvery,
 	}
 
 	o.initPools()
@@ -97,6 +100,16 @@ func (o *options) SetCapacity(value int) Options {
 
 func (o *options) Capacity() int {
 	return o.capacity
+}
+
+func (o *options) SetFlushEvery(value int) Options {
+	opts := *o
+	opts.flushEvery = value
+	return &opts
+}
+
+func (o *options) FlushEvery() int {
+	return o.flushEvery
 }
 
 func (o *options) SetStreamPool(value StreamPool) Options {
--- a/aggregation/quantile/cm/options.go
+++ b/aggregation/quantile/cm/options.go
@@ -44,7 +44,6 @@ var (
 
 	errInvalidEps       = fmt.Errorf("epsilon value must be between %f and %f", minEps, maxEps)
 	errInvalidQuantiles = fmt.Errorf("quantiles must be nonempty and between %f and %f", minQuantile, maxQuantile)
-	errNoSamplePool     = errors.New("no sample pool set")
 	errNoFloatsPool     = errors.New("no floats pool set")
 	errNoStreamPool     = errors.New("no stream pool set")
 )
@@ -140,9 +139,6 @@ func (o *options) Validate() error {
 	if o.streamPool == nil {
 		return errNoStreamPool
 	}
-	if o.samplePool == nil {
-		return errNoSamplePool
-	}
 	if o.floatsPool == nil {
 		return errNoFloatsPool
 	}
@@ -155,9 +151,6 @@ func (o *options) Validate() error {
 }
 
 func (o *options) initPools() {
-	o.samplePool = NewSamplePool(nil)
-	o.samplePool.Init()
-
 	o.floatsPool = pool.NewFloatsPool(defaultBuckets, nil)
 	o.floatsPool.Init()
 
--- a/aggregation/quantile/cm/stream.go
+++ b/aggregation/quantile/cm/stream.go
@@ -45,13 +45,15 @@ type stream struct {
 	eps             float64         // desired epsilon for errors
 	quantiles       []float64       // sorted target quantiles
 	capacity        int             // stream capacity
+	flushEvery      int             // stream flushing frequency
 	streamPool      StreamPool      // pool of streams
 	floatsPool      pool.FloatsPool // pool of float64 slices
 	acquireSampleFn acquireSampleFn // function to acquire samples
 	releaseSampleFn releaseSampleFn // function to release samples
 
 	closed          bool       // whether the stream is closed
-	numValues       int64      // number of values
+	numAdded        int        // number of values added
+	numValues       int64      // number of values inserted into the sorted stream
 	bufLess         minHeap    // sample buffer whose value is less than that at the insertion cursor
 	bufMore         minHeap    // sample buffer whose value is more than that at the insertion cursor
 	samples         sampleList // sample list
@@ -88,6 +90,7 @@ func NewStream(opts Options) Stream {
 		eps:             opts.Eps(),
 		quantiles:       opts.Quantiles(),
 		capacity:        opts.Capacity(),
+		flushEvery:      opts.FlushEvery(),
 		streamPool:      opts.StreamPool(),
 		floatsPool:      opts.FloatsPool(),
 		acquireSampleFn: acquireSampleFn,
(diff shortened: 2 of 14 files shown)


aggregation/quantile/cm/options.go
```diff
@@ -61,9 +62,10 @@ type options struct {
 // NewOptions creates a new options
```
jeromefroe: Missing a period at the end of the comment

aggregation/quantile/cm/stream.go
```diff
@@ -34,17 +34,26 @@ var (
 	nan = math.NaN()
 )
 
+// acquireSampleFn acquires a new sample.
+type acquireSampleFn func() *Sample
+
+// releaseSampleFn releases a sample.
+type releaseSampleFn func(*Sample)
+
 // stream represents a data stream
```
jeromefroe: nit: missing a period

aggregation/quantile/cm/options.go
xichen2020: Yeah I'm gonna put up a separate PR that does a blanket sweep across the entire m3aggregator repo looking for comments missing periods.

aggregation/quantile/cm/stream.go
xichen2020: See above

aggregation/quantile/cm/stream.go
```diff
@@ -311,7 +345,9 @@ func (s *stream) addToMinHeap(heap *minHeap, value float64) {
 	// and return the current heap to the pool
 	if len(curr) == cap(curr) {
 		newHeap := s.floatsPool.Get(2 * len(curr))
-		newHeap = append(newHeap, curr...)
+		for i := 0; i < len(curr); i++ {
```
jeromefroe: Might be cheaper to use `copy` here, i.e. `copy(newHeap, curr)`.

aggregation/quantile/tdigest/tdigest.go
```diff
@@ -297,7 +297,9 @@ func (d *tDigest) upperBound(index int) float64 {
 func (d *tDigest) appendCentroid(centroids []Centroid, c Centroid) []Centroid {
 	if len(centroids) == cap(centroids) {
 		newCentroids := d.centroidsPool.Get(2 * len(centroids))
-		newCentroids = append(newCentroids, centroids...)
+		for i := 0; i < len(centroids); i++ {
```
jeromefroe: Same as above, maybe we want to use `copy` here.

services/m3aggregator/config/aggregator.go
```diff
@@ -46,6 +46,10 @@ import (
 	"github.com/spaolacci/murmur3"
 )
 
+const (
+	initialBufferSizeGrowthFactor = 2
```
jeromefroe: Do we want to make this twice the flush size? Seems like that might be overkill. As a possible alternative we could make this a function call so we could use a tighter bound. For example:

```
func (c *AggregatorConfiguration) getBufferSize(n int) int {
    // Buffer size should be 1.5 times the maximum flush size to account for overflow.
    return c.MaxFlushSize * 3 * 2
}

```

services/m3aggregator/config/aggregator.go
```diff
@@ -282,11 +286,16 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	entryPool.Init(func() *aggregator.Entry { return aggregator.NewEntry(nil, opts) })
 
 	// Set buffered encoder pool.
+	// NB(xichen): we preallocate a bit over the maximum flush size as a safety measure
```
jeromefroe: Perhaps move this line to before the line where we determine the `intialBufferSize` just so its a littler clearer that that is the line its referring to?

aggregation/quantile/cm/stream.go
xichen2020: The newly allocated slice is a zero-length slice with bigger capacity, don't think you can use `copy` here.

aggregation/quantile/tdigest/tdigest.go
xichen2020: The newly allocated slice is a zero-length slice with bigger capacity, don't think you can use `copy` here.

services/m3aggregator/config/aggregator.go
xichen2020: Why 1.5 times? I think it roughly doubles the capacity if you write past the pre-allocated capacity.

services/m3aggregator/config/aggregator.go
xichen2020: Sure thing

aggregation/quantile/cm/stream.go
jeromefroe: Aggh, yea, if the slice we get back has zero length then `copy` won't work. Any reason to not use `append(newHeap, curr...)` then? It seems like the compiler should be able to optimize that into a single copy under the hood.

services/m3aggregator/config/aggregator.go
jeromefroe: It doesn't have to be 1.5 times, that was just for illustration. I'm just thinking that if the buffer is, for example, 10 kB, it might make more sense to allocate a buffer that is 12 kB instead of 20 kB, since the the overflow will likely be less than 2kB.

xichen2020: Sure, I can make it a bit more conservative, e.g., 1.5x. The number of buffers is actually pretty small (8096) so even allocating 2x we are allocating 8MB more (25MB vs. 17MB)

aggregation/quantile/cm/stream.go
xichen2020: I don't trust the compiler to not allocate a new slice so I'd rather do it myself =p

jeromefroe: Fair enough :)

services/m3aggregator/config/aggregator.go
jeromefroe: Hmm, in that case, perhaps we should just go with the simple approach of using 2x. I was thinking that the number of buffers would be closer to the number of connections (~15-20K) in which case the savings might be more significant, especially if we ever decided to increase the size of the buffers themselves.

aggregation/quantile/cm/stream.go
cw9: I think it def allocates a new slice when you pass in curr...

aggregator/list.go
```diff
@@ -110,7 +112,7 @@ func newMetricList(resolution time.Duration, opts Options) *metricList {
 		resolution:    resolution,
 		flushInterval: flushInterval,
 		aggregations:  list.New(),
-		encoder:       msgpack.NewAggregatedEncoder(encoderPool.Get()),
+		encoder:       msgpack.NewAggregatedEncoder(encoder),
```
cw9: can we do the encoder.Reset() in encoderPool.Get()? so you don't have to call this every time.

server/msgpack/server.go
```diff
@@ -205,7 +205,14 @@ func (s *server) handleConnection(conn net.Conn) {
 	// If there is an error during decoding, it's likely due to a broken connection
 	// and therefore we ignore the EOF error
 	if err := it.Err(); err != nil && err != io.EOF {
-		s.log.Errorf("decode error: %v", err)
+		remoteAddress := "<unknown>"
```
cw9: nit: make a const for "<unknown>"?

aggregation/quantile/cm/stream.go
jeromefroe: I don't think it allocates a new slice for `curr..` but creates a subslice for it instead. I wrote some [quick benchmarks](https://github.com/jeromefroe/golang_benchmarks#append) and the `curr..` approach was actually much faster and didn't cause any allocations.

xichen2020: Yeah it doesn't allocate a new slice (see https://play.golang.org/p/MUxy9zCY9W), which is somewhat surprising in itself.

cw9: hmm, in this case we should probably use the curr.. since the test shows there is no allocation

xichen2020: Sounds good, will revert the change

aggregation/quantile/tdigest/tdigest.go
xichen2020: Will revert the change

aggregator/list.go
xichen2020: It's somewhat conventional to do `object := pool.Get(); object.Reset()` before using the pooled object, this is also how we use pooled objects in m3db.

server/msgpack/server.go
xichen2020: Sure thing

PR #20, opened 2017-06-02
Change stream to take in quantiles in ResetSetData

cw9:
Which will be used in the custom aggregation change

@xichen2020 @jeromefroe

### DIFF
--- a/aggregator/options.go
+++ b/aggregator/options.go
@@ -37,6 +37,11 @@ import (
 	"github.com/spaolacci/murmur3"
 )
 
+const (
+	minQuantile = 0.0
+	maxQuantile = 1.0
+)
+
 var (
 	defaultMetricPrefix           = []byte("stats.")
 	defaultCounterPrefix          = []byte("counts.")
@@ -60,6 +65,9 @@ var (
 		policy.NewPolicy(10*time.Second, xtime.Second, 2*24*time.Hour),
 		policy.NewPolicy(time.Minute, xtime.Minute, 40*24*time.Hour),
 	}
+	defaultTimerQuantiles = []float64{0.5, 0.95, 0.99}
+
+	errInvalidQuantiles = fmt.Errorf("quantiles must be nonempty and between %f and %f", minQuantile, maxQuantile)
 )
 
 type options struct {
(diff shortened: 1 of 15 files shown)


aggregator/options.go
```diff
@@ -494,6 +514,13 @@ func (o *options) FullGaugePrefix() []byte {
 	return o.fullGaugePrefix
 }
 
+func (o *options) SetTimerQuantiles(quantiles []float64) Options {
+	opts := *o
+	opts.timerQuantiles = quantiles
+	o.computeTimerQuantilesAndSuffixes()
```
xichen2020: Should this be `opts.computeTimerQuantilesAndSuffixes()`?

aggregator/options.go
```diff
@@ -542,7 +569,6 @@ func (o *options) computeFullPrefixes() {
 }
 
 func (o *options) computeTimerQuantilesAndSuffixes() {
```
xichen2020: Should this be called `computeTimerQuantileSuffixes` then?

aggregator/types.go
```diff
@@ -336,6 +339,9 @@ type Options interface {
 	// FullGaugePrefix returns the full prefix for gauges
 	FullGaugePrefix() []byte
 
+	// SetTimerQuantiles sets the timer quantiles
+	SetTimerQuantiles(quantiles []float64) Options
```
xichen2020: These should probably be moved up to the "Read-write" base options part?

services/m3aggregator/config/aggregator.go
```diff
@@ -176,6 +179,7 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	}
 	opts = opts.SetTimerQuantileSuffixFn(quantileSuffixFn)
 
+	opts = opts.SetTimerQuantiles(c.TimerQuantiles)
```
xichen2020: nit: can chain this up with setting the suffix fn

aggregator/options.go
cw9: good catch

aggregator/options.go
cw9: sg

aggregator/types.go
cw9: sg

services/m3aggregator/config/aggregator.go
cw9: sg

PR #21, opened 2017-06-05
More performance optimizations

xichen2020:
cc @cw9 @jeromefroe @prateek @dgromov 

This PR adds more performance optimizations for the agg tier, including:
* Make elem lock per time bucket. Previously we used to lock the entire element during writing or flushing, regardless of which time bucket we are writing too, resulting in higher lock contention than necessary. This PR now makes the lock per time bucket and thus more granular, so that writes to the current time bucket don't contend with reads from the previous time bucket during flushing.
* Add option to support periodic insertion and compression. For high frequency timers in large timer batches, it is very important to add timers as fast as possible. By only performing insertion and compression, I noticed a p99 write latency drop from 400ms to 30ms for higher frequency timer writes.
* Add option to support splitting large timer batches into smaller ones. This provides the flexibility to write a large batch of timers in smaller batches, therefore making each batched write hold locks for shorter period of time and thus reduce lock contention with flushes. I noticed significant flush time reduction for high frequency timers in large batches due to batch splitting (8s -> 1s).

### DIFF
--- a/aggregator/elem.go
+++ b/aggregator/elem.go
@@ -81,8 +81,23 @@ type metricElem interface {
 	Close()
 }
 
+type timedCounter struct {
+	timeNanos int64
+	counter   *aggregation.LockedCounter
+}
+
+type timedTimer struct {
+	timeNanos int64
+	timer     *aggregation.LockedTimer
+}
+
+type timedGauge struct {
+	timeNanos int64
+	gauge     *aggregation.LockedGauge
+}
+
 type elemBase struct {
-	sync.Mutex
+	sync.RWMutex
 
 	opts       Options
 	id         id.RawID
@@ -91,40 +106,28 @@ type elemBase struct {
 	closed     bool
 }
 
-type timedCounter struct {
-	timeNanos int64
-	counter   aggregation.Counter
-}
-
-type timedTimer struct {
-	timeNanos int64
-	timer     aggregation.Timer
-}
-
-type timedGauge struct {
-	timeNanos int64
-	gauge     aggregation.Gauge
-}
-
 // CounterElem is the counter element
 type CounterElem struct {
 	elemBase
 
-	values []timedCounter // aggregated counters sorted by time in ascending order
+	values    []timedCounter // aggregated counters sorted by time in ascending order
+	toConsume []timedCounter
 }
 
 // TimerElem is the timer element
 type TimerElem struct {
 	elemBase
 
-	values []timedTimer // aggregated timers sorted by time in ascending order
+	values    []timedTimer // aggregated timers sorted by time in ascending order
+	toConsume []timedTimer
 }
 
 // GaugeElem is the gauge element
 type GaugeElem struct {
 	elemBase
 
-	values []timedGauge // aggregated gauges sorted by time in ascending order
+	values    []timedGauge // aggregated gauges sorted by time in ascending order
+	toConsume []timedGauge
 }
 
 // ResetSetData resets the counter and sets data.
@@ -137,9 +140,9 @@ func (e *elemBase) ResetSetData(id id.RawID, policy policy.Policy) {
 
 // ID returns the metric id.
 func (e *elemBase) ID() id.RawID {
-	e.Lock()
+	e.RLock()
 	id := e.id
-	e.Unlock()
+	e.RUnlock()
 	return id
 }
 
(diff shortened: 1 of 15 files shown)


aggregator/elem.go
```diff
@@ -192,17 +194,27 @@ func (e *CounterElem) Consume(earlierThanNanos int64, fn aggMetricFn) bool {
...
 		e.values = e.values[:n]
 	}
 	canCollect := len(e.values) == 0 && e.tombstoned
 	e.Unlock()
+
+	// Process the counters.
+	for i := range e.toConsume {
+		endAtNanos := e.toConsume[i].timeNanos + int64(e.policy.Resolution().Window)
+		e.toConsume[i].counter.Lock()
+		e.processValue(endAtNanos, e.toConsume[i].counter.Counter, fn)
```
jeromefroe: Do we need to lock the counter when we call `processValue`? Just thinking that a possible alternative would be to use an atomic int64 and do atomic loads and increments. It seems `processValue` just calls  `fn` with the counter's current value so if we're okay with allowing updates to the counter after we load its value we won't need a lock.

aggregator/elem.go
```diff
@@ -447,17 +534,26 @@ func (e *GaugeElem) Consume(earlierThanNanos int64, fn aggMetricFn) bool {
...
+		e.toConsume = append(e.toConsume, e.values[:idx]...)
 		n := copy(e.values[0:], e.values[idx:])
 		e.values = e.values[:n]
 	}
 	canCollect := len(e.values) == 0 && e.tombstoned
 	e.Unlock()
+
+	for i := range e.toConsume {
+		endAtNanos := e.toConsume[i].timeNanos + int64(e.policy.Resolution().Window)
+		e.toConsume[i].gauge.Lock()
```
jeromefroe: Same comment as for counters, is it possible to use an atomic variable here? Gauges are a little trickier since they're floats but we could store the value as a uint64 and then convert to and from a float when we get and set its value. I'm not familiar with the requirements of the aggregation function `fn` so it may not be viable, figured I would suggest it just in case though.

aggregator/entry.go
```diff
@@ -112,6 +112,53 @@ func (e *Entry) AddMetricWithPoliciesList(
...
+	}
+}
+
+func (e *Entry) writeBatchTimerWithPoliciesList(
+	mu unaggregated.MetricUnion,
+	pl policy.PoliciesList,
+) error {
+	// If there is no limit on the maximum batch size per write, write
+	// all timers at once.
+	maxTimerBatchSizePerWrite := e.opts.MaxTimerBatchSizePerWrite()
```
jeromefroe: This optimization seems a little unintuitive to me because we're going to have to acquire the locks we need multiple times if the number of timer values is greater than the max batch timer size. So even though other calls to `AddMetricWithPoliciesList` may have to wait less time to acquire the locks they need, calls with large batches will take longer since they have to acquire the locks multiple times. I suppose there might an equilibrium point for the percentage of calls with large batches below which it makes sense to have a max batch timer size and above which it does not. It might be interesting to do an experiment with the agg tier where we set half the cluster to use a max batch timer size and the other half to set it to zero for an extended period of time so we can compare the two.

aggregator/elem.go
xichen2020: Unfortunately we'll need it when we support custom aggregation functions (min/max/avg/etc)

aggregator/elem.go
xichen2020: See above

aggregator/entry.go
xichen2020: This is primarily to reduce the locking overhead observed when list needs to flush and acquire a write time lock shared by every entry. For a large batch size, it holds the lock for extended period of time and the list cannot flush before that lock is released. By splitting the batches into smaller ones, it gives the flushing list more chances to get the time lock, therefore reducing the average time to acquire the time lock.

services/m3aggregator/config/aggregator.go
```diff
@@ -135,6 +135,9 @@ type AggregatorConfiguration struct {
 	// EntryCheckBatchPercent determines the percentage of entries checked in a batch.
 	EntryCheckBatchPercent float64 `yaml:"entryCheckBatchPercent" validate:"min=0.0,max=1.0"`
 
+	// MaxTimerBatchSizePerWrite determines the maximum timer batch size for each batched write.
+	MaxTimerBatchSizePerWrite int `yaml:"maxTimerBatchSizePerWrite"`
```
cw9: validate min:0?

aggregator/entry.go
```diff
@@ -112,6 +112,53 @@ func (e *Entry) AddMetricWithPoliciesList(
 	mu unaggregated.MetricUnion,
 	pl policy.PoliciesList,
 ) error {
+	switch mu.Type {
+	case unaggregated.BatchTimerType:
+		err := e.writeBatchTimerWithPoliciesList(mu, pl)
+		if mu.BatchTimerVal != nil && mu.TimerValPool != nil {
```
xichen2020: Well then I'll need to write it out twice, I guess I could

services/m3aggregator/config/aggregator.go
xichen2020: Sure thing

PR #25, opened 2017-06-09
Add a broadcast flush handler

cw9:
@xichen2020

### DIFF
--- a/aggregator/handler/broadcast.go
+++ b/aggregator/handler/broadcast.go
@@ -21,77 +21,34 @@
 package handler
 
 import (
-	"bytes"
-	"sync/atomic"
-
 	"github.com/m3db/m3aggregator/aggregator"
-	"github.com/m3db/m3metrics/protocol/msgpack"
+	"github.com/m3db/m3x/errors"
 )
 
 type broadcastHandler struct {
-	handlers    []aggregator.Handler
-	handlersLen int32
+	handlers []aggregator.Handler
 }
 
 // NewBroadcastHandler creates a new Handler that routes a
 // msgpack buffer to a list of handlers.
 func NewBroadcastHandler(handlers []aggregator.Handler) aggregator.Handler {
-	return &broadcastHandler{handlers: handlers, handlersLen: int32(len(handlers))}
+	return &broadcastHandler{handlers: handlers}
 }
 
-func (b *broadcastHandler) Handle(buffer msgpack.Buffer) error {
-	var (
-		buf = newReferencedBuffer(buffer)
-		err error
-	)
+func (b *broadcastHandler) Handle(buffer *aggregator.RefCountedBuffer) error {
+	multiErr := xerrors.NewMultiError()
 	for _, h := range b.handlers {
-		buf.incRef()
-		if err = h.Handle(buf); err != nil {
-			break
+		buffer.IncRef()
+		if err := h.Handle(buffer); err != nil {
+			multiErr = multiErr.Add(err)
 		}
 	}
-	buf.Close()
-	return err
+	buffer.DecRef()
+	return multiErr.FinalError()
 }
 
 func (b *broadcastHandler) Close() {
 	for _, h := range b.handlers {
 		h.Close()
 	}
-}
-
-type referencedBuffer struct {
-	// Could not embed msgpack.Buffer here due to naming conflict with method Buffer().
-	buffer msgpack.Buffer
-	ref    uint32
-}
-
-func newReferencedBuffer(buffer msgpack.Buffer) *referencedBuffer {
-	return &referencedBuffer{buffer: buffer, ref: 1}
-}
-
-func (b *referencedBuffer) Buffer() *bytes.Buffer {
-	return b.buffer.Buffer()
-}
-
-func (b *referencedBuffer) Bytes() []byte {
-	return b.buffer.Bytes()
-}
-
-func (b *referencedBuffer) Reset() {
-	b.buffer.Reset()
-}
-
-func (b *referencedBuffer) Close() {
-	// Decrement b.ref
-	ref := atomic.AddUint32(&b.ref, ^uint32(0))
-	if ref > 0 {
-		return
-	}
-
-	b.buffer.Close()
-}
-
-func (b *referencedBuffer) incRef() {
-	atomic.AddUint32(&b.ref, 1)
 }
--- a/aggregator/handler/broadcast.go
+++ b/aggregator/handler/broadcast.go
@@ -1,0 +1,97 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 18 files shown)


services/m3aggregator/config/aggregator.go
```diff
@@ -124,7 +125,7 @@ type AggregatorConfiguration struct {
 	MaxFlushSize int `yaml:"maxFlushSize"`
 
 	// Flushing handler configuration.
-	Flush flushHandlerConfiguration `yaml:"flush"`
+	Flush []flushHandlerConfiguration `yaml:"flush"`
```
xichen2020: I think we should keep this as `flushHandlerConfiguration` and define a separate `broadcastHandlerConfiguration` which contains a list of `flushHandlerConfiguration`, i.e., the config will look like:
```
flush:
    type: broadcast
    broadcast:
      - type: forward
         forward:
           ...
       - type: forward
           ...
```
This makes it more flexible to configure the handlers, and doesn't force the users to always broadcast if they only have a single handler.

aggregator/handler/broadcast.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (b *broadcastHandler) Handle(buffer msgpack.Buffer) error {
+	var (
+		buf = newReferencedBuffer(buffer)
+		err error
+	)
+	for _, h := range b.handlers {
+		buf.incRef()
+		if err = h.Handle(buf); err != nil {
```
xichen2020: Should we continue even if we encounter an error? Otherwise one badly behaving handler could prevent all other handlers handling values.

Also we can just pass a refcounted buffer into the `Handle` method, and `decRef` will just call a destructor (e.g., closing the buffer) when the ref count goes to zero.

aggregator/handler/broadcast.go
```diff
@@ -0,0 +1,97 @@
...
+func (b *broadcastHandler) Close() {
+	for _, h := range b.handlers {
+		h.Close()
+	}
+}
+
+type referencedBuffer struct {
+	// Could not embed msgpack.Buffer here due to naming conflict with method Buffer().
+	buffer msgpack.Buffer
+	ref    uint32
```
xichen2020: Can we make this an `int`?

aggregator/handler/config.go
```diff
@@ -0,0 +1,169 @@
...
+// FlushHandlerConfiguration contains configuration for flushing metrics.
+type FlushHandlerConfiguration struct {
+	// Flushing handler type.
+	Type string `yaml:"type" validate:"regexp=(^blackhole$|^logging$|^forward$|^broadcast$)"`
+
+	// Forward handler configuration.
+	Forward *forwardHandlerConfiguration `yaml:"forward"`
+
+	// Forward handler configuration.
+	Broadcast []broadcastHandlerConfiguration `yaml:"broadcast"`
```
xichen2020: This should be `validate:"nonzero"`?

aggregator/handler/config.go
```diff
@@ -0,0 +1,169 @@
...
+		h, err := cfg.NewHandler(iopts)
+		if err != nil {
+			return nil, err
+		}
+		handlers[i] = h
+	}
+	return NewBroadcastHandler(handlers), nil
+}
+
+type broadcastHandlerConfiguration struct {
```
xichen2020: Shouldn't this just be `FlushHandlerConfiguration`?

aggregator/handler/forward_options.go
```diff
@@ -99,7 +99,7 @@ func NewForwardHandlerOptions() ForwardHandlerOptions {
 		connectTimeout:         defaultConnectTimeout,
 		connectionKeepAlive:    defaultConnectionKeepAlive,
 		connectionWriteTimeout: defaultConnectionWriteTimeout,
-		reconnectRetrier:       xretry.NewRetrier(xretry.NewOptions()),
+		reconnectRetrier:       xretry.NewRetrier(xretry.NewOptions().SetForever(true)),
```
xichen2020: Perhaps we should have a `NewForeverRetrier` ctor that does this?

aggregator/handler/handler.go
```diff
@@ -28,4 +28,5 @@ const (
 	BlackholeHandler Type = "blackhole"
 	LoggingHandler   Type = "logging"
 	ForwardHandler   Type = "forward"
+	Broadcast        Type = "broadcast"
```
xichen2020: `BroadcastHandler` to be consistent?

aggregator/handler/config.go
```diff
@@ -0,0 +1,153 @@
...
+			"forward-target": c.Forward.Name,
+		})
+		logger := iOpts.Logger().WithFields(xlog.NewLogField("forward-target", c.Forward.Name))
+		return c.Forward.NewHandler(iOpts.SetMetricsScope(scope).SetLogger(logger))
+	case Broadcast:
+		if c.Broadcast == nil {
+			return nil, errNoBroadcastHandlerConfiguration
+		}
+		scope = scope.SubScope("broadcast").Tagged(map[string]string{
+			"handler-type": "broadcast",
```
cw9: the value for tag "handler-type" will always be override by the sub handler under the broadcast handler, and the metrics name will be "broadcast.forward.foo", which breaks the current metrics. This is why I separated the configuration in my original code, so when constructing the forward handler under a broadcast handler, it only sets "target" tag but not override the "handler-type" tag. Then we can have "target": "ingester", "handler-type": "broadcast".

We also probably don't want to broadcast to a broadcast handler recursively?

aggregator/handler/handler.go
cw9: Should be BroadcastHandler

aggregator/ref_count_buffer.go
```diff
@@ -0,0 +1,62 @@
...
+// NewRefCountedBuffer creates a new refcounted buffer.
+func NewRefCountedBuffer(buffer msgpack.Buffer) *RefCountedBuffer {
+	return &RefCountedBuffer{n: 1, buf: buffer}
+}
+
+// Buffer returns the internal msgpack buffer.
+func (b *RefCountedBuffer) Buffer() msgpack.Buffer { return b.buf }
+
+// IncRef increments the refcount for the buffer.
+func (b *RefCountedBuffer) IncRef() int {
```
cw9: why not just return int32?

aggregator/ref_count_buffer.go
```diff
@@ -0,0 +1,62 @@
...
+	if n := int(atomic.AddInt32(&b.n, 1)); n > 0 {
+		return n
+	}
+	panic("invalid ref count")
+}
+
+// DecRef decrements the refcount for the buffer.
+func (b *RefCountedBuffer) DecRef() int {
```
cw9: why not just return int32?

aggregator/handler/config.go
xichen2020: Yeah I don't think we can use tag to express hierarchy, which is why I used subscope to express the follow of metrics (i.e., flowing from broadcast handler to forward handler to whatever). I don't think there's anything wrong with "broadcast.forward.foo", this is a new metric we'll need to set up anyway.

I think it'll be more flexible to give users options to configure the handlers in any way they want to. There is no reason why broadcast handler should be treated any differently from other handler types since it is simply another type of handler.

aggregator/ref_count_buffer.go
xichen2020: Seems like an implementation detail though, caller shouldn't care about whether we use int32 or uint32 or int64 internally to track the ref. The return type should be the general type used to count things (number of references in this case), which IMO should be int because that's what we use for `len` too.

aggregator/ref_count_buffer.go
xichen2020: See below, actually in this case I don't even need to return the int, so I'll just make it a void function.

PR #29, opened 2017-06-15
Fix config, make sure timer flush fast path is working

cw9:
@xichen2020

### DIFF
--- a/aggregator/elem.go
+++ b/aggregator/elem.go
@@ -362,6 +362,8 @@ func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes 
 	e.elemBase.ResetSetData(id, sp, aggTypes)
 
 	if aggTypes.IsDefault() {
+		// TODO(cw) Remove UseDefaultAggregation in aggregation opts and
+		// store it in elemBase.
 		e.aggOpts.ResetSetData(e.opts.DefaultTimerAggregationTypes())
 		e.isAggTypesPooled = false
 		e.quantiles, e.isQuantilesPooled = e.opts.TimerQuantiles(), false
--- a/aggregator/elem.go
+++ b/aggregator/elem.go
@@ -359,16 +359,16 @@ func NewTimerElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggregat
 
 // ResetSetData overrides elemBase.ResetData, to include quantile update for timer elements.
 func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
+	e.elemBase.ResetSetData(id, sp, aggTypes)
+
 	if aggTypes.IsDefault() {
-		aggTypes = e.opts.DefaultTimerAggregationTypes()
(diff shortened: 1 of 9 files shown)


aggregator/elem.go
```diff
@@ -359,16 +359,16 @@ func NewTimerElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggregat
 
 // ResetSetData overrides elemBase.ResetData, to include quantile update for timer elements.
 func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
+	e.elemBase.ResetSetData(id, sp, aggTypes)
+
 	if aggTypes.IsDefault() {
-		aggTypes = e.opts.DefaultTimerAggregationTypes()
+		e.aggOpts.ResetSetData(e.opts.DefaultTimerAggregationTypes())
```
xichen2020: Put a TODO in there to ensure the aggTypes stored in elemBase agree with what we store in aggregation options?

config/m3aggregator.yaml
```diff
@@ -76,10 +77,6 @@ aggregator:
           capacity: 32
         - count: 1024
           capacity: 64
-  timerQuantiles:
```
xichen2020: Can we keep the timer quantiles? I'd rather keep them explicit if possible

config/m3aggregator.yaml
```diff
@@ -61,6 +61,7 @@ http:
   writeTimeout: 45s
 
 aggregator:
+  timerAggregationTypes: Sum,SumSq,Mean,Lower,Upper,Count,Stdev,Median,P50,P95,P99
```
xichen2020: Add a comment here saying they need to be comma deliminated string? Might be more consistent to make it a list like the quantiles, though

aggregator/elem.go
cw9: Sure

config/m3aggregator.yaml
cw9: Sure, we could change it to a list of agg types eventually

config/m3aggregator.yaml
cw9: We can't, as we don't support setting timerquantiles in the options now, and quantiles needs to be valid aggregation types anyway so we should just set aggregation types

config/m3aggregator.yaml
xichen2020: Yeah can we do that in this diff? A list of types looks much easier on the eyes

config/m3aggregator.yaml
xichen2020: Right yeah makes sense

xichen2020:
Also do we have a regression test that catches the bug you fixed? If not, can we add it?

cw9:
I added a regression test in elem_test.go

cw9:
You need to stamp this diff for the yaml parsing to work: https://github.com/m3db/m3metrics/pull/53

xichen2020:
Yeah stamped the other diff.

config/m3aggregator.yaml
cw9: done

PR #32, opened 2017-06-18
Add default aggregation types for Counter and Gauge

cw9:
Reordered functions in aggregator.Options.
Allow suffix override through config per metrics type per aggregation type.

@xichen2020

### DIFF
--- a/aggregator/elem.go
+++ b/aggregator/elem.go
@@ -144,11 +144,11 @@ func newElemBase(opts Options) elemBase {
 }
 
 // resetSetData resets the element base and sets data.
-func (e *elemBase) resetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes, isDefault bool) {
+func (e *elemBase) resetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes, useDefaultAggregation bool) {
 	e.id = id
 	e.sp = sp
 	e.aggTypes = aggTypes
-	e.useDefaultAggregation = isDefault
+	e.useDefaultAggregation = useDefaultAggregation
 	e.aggOpts.ResetSetData(aggTypes)
 	e.tombstoned = false
 	e.closed = false
@@ -186,12 +186,12 @@ func NewCounterElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggreg
 
 // ResetSetData resets the counter element and sets data.
 func (e *CounterElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
-	isDefault := aggTypes.IsDefault()
-	if isDefault {
+	useDefaultAggregation := aggTypes.IsDefault()
+	if useDefaultAggregation {
 		aggTypes = e.opts.DefaultCounterAggregationTypes()
 	}
 
-	e.elemBase.resetSetData(id, sp, aggTypes, isDefault)
+	e.elemBase.resetSetData(id, sp, aggTypes, useDefaultAggregation)
 }
 
 // AddMetric adds a new counter value
@@ -366,15 +366,15 @@ func NewTimerElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggregat
 
 // ResetSetData resets the timer element and sets data.
 func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
-	isDefault := aggTypes.IsDefault()
-	if isDefault {
+	useDefaultAggregation := aggTypes.IsDefault()
+	if useDefaultAggregation {
 		aggTypes = e.opts.DefaultTimerAggregationTypes()
 		e.quantiles, e.isQuantilesPooled = e.opts.TimerQuantiles(), false
 	} else {
 		e.quantiles, e.isQuantilesPooled = aggTypes.PooledQuantiles(e.opts.QuantilesPool())
 	}
 
-	e.elemBase.resetSetData(id, sp, aggTypes, isDefault)
+	e.elemBase.resetSetData(id, sp, aggTypes, useDefaultAggregation)
 }
 
 // AddMetric adds a new batch of timer values.
@@ -565,12 +565,12 @@ func NewGaugeElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggregat
 
 // ResetSetData resets the gauge element and sets data.
 func (e *GaugeElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
-	isDefault := aggTypes.IsDefault()
-	if isDefault {
+	useDefaultAggregation := aggTypes.IsDefault()
+	if useDefaultAggregation {
 		aggTypes = e.opts.DefaultGaugeAggregationTypes()
 	}
 
-	e.elemBase.resetSetData(id, sp, aggTypes, isDefault)
+	e.elemBase.resetSetData(id, sp, aggTypes, useDefaultAggregation)
 }
 
 // AddMetric adds a new gauge value
--- a/aggregator/elem.go
+++ b/aggregator/elem.go
@@ -99,14 +99,14 @@ type timedGauge struct {
 type elemBase struct {
 	sync.RWMutex
 
-	opts       Options
-	id         id.RawID
-	sp         policy.StoragePolicy
-	aggTypes   policy.AggregationTypes
-	aggOpts    aggregation.Options
-	isDefault  bool
-	tombstoned bool
-	closed     bool
+	opts                  Options
+	id                    id.RawID
+	sp                    policy.StoragePolicy
+	aggTypes              policy.AggregationTypes
+	aggOpts               aggregation.Options
+	useDefaultAggregation bool
+	tombstoned            bool
+	closed                bool
 }
 
 // CounterElem is the counter element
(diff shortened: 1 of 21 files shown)


aggregator/elem.go
```diff
@@ -104,6 +104,7 @@ type elemBase struct {
 	sp         policy.StoragePolicy
 	aggTypes   policy.AggregationTypes
 	aggOpts    aggregation.Options
+	isDefault  bool
```
xichen2020: `useDefaultAggTypes` is more clear?

aggregator/elem.go
```diff
@@ -357,20 +364,17 @@ func NewTimerElem(id id.RawID, sp policy.StoragePolicy, aggTypes policy.Aggregat
...
+// ResetSetData resets the timer element and sets data.
 func (e *TimerElem) ResetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes policy.AggregationTypes) {
-	e.elemBase.ResetSetData(id, sp, aggTypes)
-
-	if aggTypes.IsDefault() {
-		// TODO(cw) Remove UseDefaultAggregation in aggregation opts and
-		// store it in elemBase.
-		e.aggOpts.ResetSetData(e.opts.DefaultTimerAggregationTypes())
-		e.isAggTypesPooled = false
+	isDefault := aggTypes.IsDefault()
```
xichen2020: Yeah I think `useDefaultAggTypes` is more intuitive, can we replace `isDefault` with `useDefaultAggTypes` wherever applicable?

aggregator/options.go
```diff
@@ -636,42 +735,89 @@ func (o *options) computeQuantiles() {
...
 			q, ok := aggType.Quantile()
 			if ok {
-				o.suffixes[aggType.ID()] = o.timerQuantileSuffixFn(q)
+				o.defaultSuffixes[aggType.ID()] = o.timerQuantileSuffixFn(q)
 			}
 		}
 	}
 }
 
+func (o *options) computeCounterSuffix() {
```
xichen2020: `computeCounterSuffixes` to be consistent? Also the three `computeXXXSuffixes` functions can probably be refactored into a common one `func computeSuffixes(suffixes *[][]byte, suffixOverride map[policy.AggregationType][]byte)`?

aggregator/options.go
```diff
@@ -622,8 +716,13 @@ func (o *options) initPools() {
 func (o *options) computeAllDerived() {
 	o.computeFullPrefixes()
 	o.computeQuantiles()
-	o.computeSuffixes()
+	o.computeDefaultSuffixes()
```
xichen2020: Add a comment here indicating the order of computing the suffixes matter.

aggregator/options.go
```diff
@@ -556,6 +613,27 @@ func (o *options) SetQuantilesPool(pool pool.FloatsPool) Options {
 	return &opts
 }
 
+func (o *options) SetCounterSuffixOverride(m map[policy.AggregationType][]byte) Options {
+	opts := *o
+	opts.counterSuffixOverride = m
+	opts.computeCounterSuffix()
```
xichen2020: Do we need to compute default counter suffixes here as well?

aggregator/options.go
```diff
@@ -556,6 +613,27 @@ func (o *options) SetQuantilesPool(pool pool.FloatsPool) Options {
...
+	return &opts
+}
+
+func (o *options) SetTimerSuffixOverride(m map[policy.AggregationType][]byte) Options {
+	opts := *o
+	opts.timerSuffixOverride = m
+	opts.computeTimerSuffix()
```
xichen2020: Do we need to compute default timer suffixes here as well?

aggregator/options.go
```diff
@@ -556,6 +613,27 @@ func (o *options) SetQuantilesPool(pool pool.FloatsPool) Options {
...
+	return &opts
+}
+
+func (o *options) SetGaugeSuffixOverride(m map[policy.AggregationType][]byte) Options {
+	opts := *o
+	opts.gaugeSuffixOverride = m
```
xichen2020: Do we need to compute default gauge suffixes here as well?

aggregator/options.go
```diff
@@ -275,42 +350,35 @@ func (o *options) AggregationCountSuffix() []byte {
...
+	return o.sumSqSuffix
 }
 
-func (o *options) SetDefaultTimerAggregationTypes(aggTypes policy.AggregationTypes) Options {
+func (o *options) SetAggregationStdevSuffix(value []byte) Options {
 	opts := *o
-	opts.defaultTimerAggregationTypes = aggTypes
-	opts.computeQuantiles()
-	opts.computeDefaultTimerAggregationSuffix()
+	opts.stdevSuffix = value
```
xichen2020: When setting each of these suffixes, don't we need to recompute the default suffixes and all the suffixes for all metric types, since it changes the default suffixes and therefore potentially the default type-specific suffixes etc?

If so, you might wanna move all the suffix computing functions into a single one and call it from each of the individual suffix setters.

aggregator/options.go
```diff
@@ -47,8 +47,8 @@ var (
 	defaultAggregationSumSuffix      = []byte(".sum")
 	defaultAggregationSumSqSuffix    = []byte(".sum_sq")
 	defaultAggregationMeanSuffix     = []byte(".mean")
-	defaultAggregationLowerSuffix    = []byte(".lower")
-	defaultAggregationUpperSuffix    = []byte(".upper")
+	defaultAggregationMinSuffix      = []byte(".min")
```
xichen2020: I'm still not sure about this, if a user uses Min for counters or gauges, he/she will get .min, but if the user uses it for timers, he/she will get ".lower", isn't it confusing?

aggregator/types.go
```diff
@@ -129,6 +129,24 @@ type Handler interface {
 type Options interface {
 	/// Read-write base options
 
+	// SetDefaultCounterAggregationTypes sets the counter aggregation types.
+	SetDefaultCounterAggregationTypes(aggTypes policy.AggregationTypes) Options
```
xichen2020: nit: the convention is simply name the parameter `value` regardless of what the actual type is.

aggregator/types.go
```diff
@@ -357,7 +369,18 @@ type Options interface {
 	// QuantilesPool returns the timer quantiles pool.
 	QuantilesPool() pool.FloatsPool
 
-	/// Read-only derived options
+	/// Write-only options.
+
+	// SetCounterSuffixOverride sets the overrides for counter suffixes.
+	SetCounterSuffixOverride(m map[policy.AggregationType][]byte) Options
```
xichen2020: Same, we typically just name it `value`.

services/m3aggregator/config/aggregator.go
```diff
@@ -171,6 +186,10 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	scope := instrumentOpts.MetricsScope()
 	opts := aggregator.NewOptions().SetInstrumentOptions(instrumentOpts)
 
+	opts = opts.SetDefaultCounterAggregationTypes(c.CounterAggregationTypes)
```
xichen2020: nit: chain these?

services/m3aggregator/config/aggregator.go
```diff
@@ -179,21 +198,24 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
...
 
 	// Set timer quantiles and quantile suffix function.
-	opts = opts.SetDefaultTimerAggregationTypes(c.TimerAggregationTypes)
 	quantileSuffixFn, err := c.parseTimerQuantileSuffixFn(opts)
 	if err != nil {
 		return nil, err
 	}
 	opts = opts.SetTimerQuantileSuffixFn(quantileSuffixFn)
 
+	opts = opts.SetCounterSuffixOverride(c.parseSuffixOverride(c.CounterSuffixOverride))
```
xichen2020: nit: chain these?

aggregator/elem.go
cw9: sure

aggregator/elem.go
cw9: sg

aggregator/options.go
cw9: I think it's actually ok, I feel lower or upper for a gauge is actually more confusing, I had to do lower and upper for timers for backwards compatibility but since we could decouple the suffix with the aggregation types, I feel we should support new metrics with less hacky suffixes

aggregator/options.go
cw9: Yeah good call, will do

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sure

aggregator/options.go
cw9: sg

aggregator/options.go
cw9: sure

aggregator/types.go
cw9: sg

aggregator/options.go
xichen2020: Yeah I guess that's fine for now, if users find them counter-intuitive, can change it back

aggregator/elem.go
```diff
@@ -148,7 +148,7 @@ func (e *elemBase) resetSetData(id id.RawID, sp policy.StoragePolicy, aggTypes p
 	e.id = id
 	e.sp = sp
 	e.aggTypes = aggTypes
-	e.isDefault = isDefault
+	e.useDefaultAggregation = isDefault
```
xichen2020: rename `isDefault` to `useDefaultAggregation`?

PR #35, opened 2017-06-27
Provide keepAlive options for the aggregation server

### LINKED ISSUE #34: Aggregation server needs to actively close connections after they are closed client-side
cc @cw9 

It appears in some cases after the client closed the connections, the servers are unaware that the connections had been closed and still wait on the connection to read more data. This means the server keeps the dead connections around and drives up the goroutine count and memory usage artificially.

xichen2020:
cc @cw9 @prateek @jeromefroe 

This PR adds keepAlive options for the aggregation server to actively close out connections that have been closed by the client.

Fixes #34 .

### DIFF
--- a/server/msgpack/options.go
+++ b/server/msgpack/options.go
@@ -21,11 +21,9 @@
 package msgpack
 
 import (
-	"time"
-
 	"github.com/m3db/m3metrics/protocol/msgpack"
 	"github.com/m3db/m3x/instrument"
-	"github.com/m3db/m3x/retry"
+	"github.com/m3db/m3x/server"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


server/msgpack/options.go
```diff
@@ -34,6 +41,18 @@ type Options interface {
 	// InstrumentOptions returns the instrument options
 	InstrumentOptions() instrument.Options
 
+	// SetKeepAliveEnabled sets the keep alive state for tcp connections.
+	SetKeepAliveEnabled(value bool) Options
```
cw9: Can we add Get/Set functions for server.Options instead?

xichen2020: Sure

services/m3aggregator/config/server.go
```diff
@@ -49,9 +56,17 @@ func (c *MsgpackServerConfiguration) NewMsgpackServerOptions(
 ) msgpack.Options {
 	opts := msgpack.NewOptions().SetInstrumentOptions(instrumentOpts)
 
-	// Set retry options.
-	retryOpts := c.Retry.NewOptions(instrumentOpts.MetricsScope())
-	opts = opts.SetRetryOptions(retryOpts)
+	// Set server options.
+	serverOpts := xserver.NewOptions().
+		SetInstrumentOptions(instrumentOpts).
```
cw9: also do Tagged(map[string]string{"server": "msgpack"})?

xichen2020: The metric scoped passed in already has `msgpack-server` as part of its name, don't think we need an additional tag here.

cw9: Having a tag makes the metrics a lot easier to locate, say we can do service: xxx server: msgpack to easily find all the related metrics.

xichen2020: Okay sure

PR #36, opened 2017-07-24
Aggregation server replication

xichen2020:
cc @robskillington @martin-mao @prateek @schallert @cw9 

This PR adds replication logic to the aggregation tier to support handling rolling deployments and single node failures.


aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,279 @@
...
+	defer mgr.Unlock()
+
+	mgr.flushTimesKey = fmt.Sprintf(mgr.flushTimesKeyFmt, shardSetID)
+	go mgr.watchFlushTimes()
+}
+
+// NB(xichen): no actions needed for initializing the follower flush manager.
+func (mgr *followerFlushManager) Init(*sync.RWMutex, []*flushBucket) {}
+
+func (mgr *followerFlushManager) Flush(
```
martin-mao: Is it possible for multiple Flush calls to pile up? Ie if the interval was 1 second or sub-second and a flush took more than the aggregation period for whatever reason. If so, what is the logic? The next flush will have nothing in it's buckets?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,279 @@
...
+// NB(xichen): the follower flush manager flushes data based on the flush times
+// stored in kv and does not need to take extra actions when a new bucket is added.
+func (mgr *followerFlushManager) OnBucketAdded(int, *flushBucket) {}
+
+func (mgr *followerFlushManager) flushersFromKVUpdate(buckets []*flushBucket) []flushersGroup {
+	flushersByInterval := make([]flushersGroup, len(buckets))
+	for i, bucket := range buckets {
+		flushersByInterval[i].interval = bucket.interval
+		flushersByInterval[i].duration = bucket.duration
+		flushersByInterval[i].flushers = make([]flusherWithTime, 0, 16)
```
martin-mao: Move 16 to a const somewhere? Why 16?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,279 @@
...
+		for _, flusher := range bucket.flushers {
+			shard := flusher.Shard()
+			shardFlushTimes, exists := mgr.proto.ByShard[shard]
```
martin-mao: mgr.proto doesn't need lock?

aggregator/aggregator.go
```diff
@@ -227,18 +270,31 @@ func (agg *aggregator) AddMetricWithPoliciesList(
...
-		return errAggregatorIsNotOpenOrClosed
+		agg.Unlock()
+		return errAggregatorExpectOpen
+	}
+	if closeType == GracefulClose {
+		agg.state = aggregatorClosing
+		agg.Unlock()
+		ctx, cancel := context.WithTimeout(context.Background(), agg.gracefulCloseTimeout)
+		defer cancel()
+		if err := agg.electionManager.Resign(ctx); err != nil {
```
xichen2020: Yes the semantics of the `Resign` API on the election manager is that the replica may only resign successfully and become a follower if it can confirm the other replica has successfully taken over the ownership. If both the leader and the follower race to resign, only one of them will win and become the follower and the other replica will fail to resign and assume the leader role, until the follower is back up and ready to take over the leader role.

Additionally, we'll ensure no two replicas owning the same shardset will ever be deployed at the same time in the deployment tooling for the agg tier.

aggregator/aggregator.go
```diff
@@ -227,18 +270,31 @@ func (agg *aggregator) AddMetricWithPoliciesList(
...
+			agg.Lock()
+			agg.state = aggregatorOpen
+			agg.Unlock()
```
xichen2020: Unfortunately we can't hold the lock while resigning (which is what we get with `defer`) because resigning needs to confirm the successful transfer of leader role with kv and as a result may take a long time (by default it times out after 5 minutes) due to underlying network blip/etcd issue, causing metrics ingestion to be backed up due to the aggregator not able to acquire the read lock in `AddMetricListWithPolicies()`.

aggregator/aggregator.go
```diff
@@ -404,6 +444,29 @@ func (agg *aggregator) ownedShards() []*aggregatorShard {
 	return ownedShards
 }
 
+func (agg *aggregator) placement() (services.Placement, error) {
+	stagedPlacement, onStagedPlacementDoneFn, err := agg.placementWatcher.ActiveStagedPlacement()
```
xichen2020: Both `ActiveStagedPlacement` and `ActivePlacement` calls are thread-safe so no locks are needed here.

aggregator/flush_mgr_options.go
```diff
@@ -35,7 +37,11 @@ const (
 )
 
 var (
-	defaultWorkerPoolSize = int(math.Max(float64(runtime.NumCPU()/8), 1.0))
+	defaultWorkerPoolSize         = int(math.Max(float64(runtime.NumCPU()/8), 1.0))
```
xichen2020: I did a global search and this is the only place in m3aggregator repo where `/8` is used. I picked this as the default since it has empirically worked well for us. Users in the OSS world of course can set different values in the config file to whatever value suits them the best.

aggregator/follower_flush_mgr.go
xichen2020: The flushes are performed sequentially so they will not pile up. Specifically the follower flush manger will check for flush time updates from kv periodically (by default 1s) and only perform a flush if there is an update or sufficient time (by default 15 minutes) has passed without a flush. The leader by default persist flush times every 10s to reduce kv write volume, so if the follower takes more than 1s to finish the flush, that's totally fine because the next kv update from the leader won't come in for another 10s. Even if the follower takes more than 10s to flush, that's still okay because the follower will simply miss an intermediate kv flush times update and instead use the latest flush times for the next flush.

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,279 @@
...
+	buckets []*flushBucket,
+) (bool, time.Duration) {
+	var (
+		shouldFlush        = false
+		flushersByInterval []flushersGroup
+	)
+	// NB(xichen): a flush is triggered in the following scenarios:
+	// * The flush times persisted in kv have been updated since last flush.
+	// * Sufficient time (a.k.a. maxNoFlushDuration) has elapsed since last flush.
+	bucketLock.RLock()
```
xichen2020: The `bucketLock` is used to synchronize access to `buckets`. The `buckets` are kept in the top-level flush manager as a global registry for the lists containing aggregated metrics in different shards, and are shared between the leader flush manager and the follower flush manager for centralized access to the registered lists. As such, the bucket lock also needs to be shared between the top-level flush manager, the leader flush manager, and the follower flush manager and cannot be merged with the individual manager locks.

aggregator/follower_flush_mgr.go
xichen2020: Sure sounds good. Picked `16` as a default small size for upfront allocation because each node in prod owns roughly 15 shards.

aggregator/follower_flush_mgr.go
xichen2020: This method is protected by a write lock. I'll rename it to `flushersFromKVUpdateWithLock` for clarity.

aggregator/aggregator.go
```diff
@@ -227,18 +270,31 @@ func (agg *aggregator) AddMetricWithPoliciesList(
...
+		agg.Unlock()
+		ctx, cancel := context.WithTimeout(context.Background(), agg.gracefulCloseTimeout)
+		defer cancel()
+		if err := agg.electionManager.Resign(ctx); err != nil {
+			agg.Lock()
+			agg.state = aggregatorOpen
+			agg.Unlock()
+			return err
+		}
+		agg.Lock()
```
robskillington: Maybe to make this a tiny bit more readable:
```go
err := agg.electionManager.Resign(ctx)
agg.Lock()
if err != nil {
  agg.state = aggregatorOpen
  agg.Unlock()
  return err
}
// no need to agg.Lock here
```

aggregator/aggregator.go
```diff
@@ -404,6 +444,29 @@ func (agg *aggregator) ownedShards() []*aggregatorShard {
 	return ownedShards
 }
 
+func (agg *aggregator) placement() (services.Placement, error) {
+	stagedPlacement, onStagedPlacementDoneFn, err := agg.placementWatcher.ActiveStagedPlacement()
+	if err != nil {
+		return nil, err
+	}
+	placement, onPlacementDoneFn, err := stagedPlacement.ActivePlacement()
```
robskillington: Perhaps instead of explicitly calling, after dealing with `return nil, err` you can just above calling `ActivePlacement()` perform a `defer onStagedPlacementDoneFn()`

aggregator/election_mgr.go
```diff
@@ -0,0 +1,465 @@
...
+	mgr.changeCancelFn = cancelFn
+	mgr.changeWg.Add(1)
+	go mgr.leaderToFollower(ctx)
+}
+
+// cancelInProgressChange cancels in progress election status changes if any,
+// returning true if a change is cancelled and false otherwise.
+func (mgr *electionManager) cancelInProgressChange() bool {
+	mgr.Lock()
+	if !mgr.changeInProgress {
```
robskillington: Probably should also check that `mgr.changeCancelFn` is not nil?

aggregator/aggregator.go
```diff
@@ -227,18 +272,30 @@ func (agg *aggregator) AddMetricWithPoliciesList(
...
+	}
+}
+
 func (agg *aggregator) Close() error {
 	agg.Lock()
-	defer agg.Unlock()
-
 	if agg.state != aggregatorOpen {
-		return errAggregatorIsNotOpenOrClosed
+		agg.Unlock()
```
robskillington: Seems like this is missing from the final return statement?  Any reason to not just defer `agg.Unlock()` like we used  to previously?

If not we need to `agg.Unlock()` after the `close(...)` yeah?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,501 @@
...
+		isFollower       = false
+		stateCheckPeriod = time.Second
+	)
+
+	for {
+		select {
+		case <-ctx.Done():
+			ctxDone = true
+		default:
+			isFollower = mgr.ElectionState() == FollowerState
```
robskillington: Might it be better to have some signal for follower state instead of a sleep and check here?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}
+	return flushersByInterval
+}
+
+func (mgr *followerFlushManager) flushersFromForcedFlushWithLock(
+	buckets []*flushBucket,
+	flushBeforeNanos int64,
+) []flushersGroup {
+	flushersByInterval := make([]flushersGroup, len(buckets))
```
martin-mao: Why does this need to be done with a lock? It looks like this func is generating flushersByInterval using params, I don't see any reads of anything off mgr.

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,344 @@
...
+		select {
+		case <-flushTimesWatch.C():
+		case <-mgr.doneCh:
+			return
+		}
+
+		value := flushTimesWatch.Get()
+		mgr.Lock()
+		mgr.proto.Reset()
+		if err = value.Unmarshal(&mgr.proto); err != nil {
```
martin-mao: Nit: Can you dothe unmarshalling into something outside of lock and then set it if successful inside the lock?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,344 @@
...
+	var (
+		mgr       = t.mgr
+		wgWorkers sync.WaitGroup
+	)
+	for _, group := range t.flushersByInterval {
+		start := mgr.nowFn()
+		for _, flusherWithTime := range group.flushers {
+			flusherWithTime := flusherWithTime
+			wgWorkers.Add(1)
+			mgr.workers.Go(func() {
```
martin-mao: We ok blocking here if the workers pool is exhausted?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,344 @@
...
+	mgr                *followerFlushManager
+	flushersByInterval []flushersGroup
+}
+
+func (t *followerFlushTask) Run() {
+	var (
+		mgr       = t.mgr
+		wgWorkers sync.WaitGroup
+	)
+	for _, group := range t.flushersByInterval {
```
martin-mao: Any reason to do the groups sequentially instead of in parallel?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+	rand                *rand.Rand
+	flushTimes          flushMetadataHeap
+	flushTimesKey       string
+	lastPersistAt       time.Time
+	flushedSincePersist bool
+	flushTask           *leaderFlushTask
+	metrics             leaderFlushManagerMetrics
+}
+
+func newLeaderFlushManager(opts FlushManagerOptions) roleBasedFlushManager {
```
martin-mao: How much do the different FlushManagers differ by? If it's not much, can you combine them into one file with some toggles for leader etc?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+
+func (mgr *leaderFlushManager) enqueueBucketWithLock(
+	bucketIdx int,
+	bucket *flushBucket,
+) {
+	nowNanos := mgr.nowNanos()
+	flushInterval := bucket.interval
+	nextFlushNanos := nowNanos + int64(flushInterval)
+	if mgr.jitterEnabled {
+		jitterNanos := int64(mgr.rand.Int63n(int64(flushInterval)))
```
martin-mao: We're going to jitter all the way up to the bucket interval? What's the max Bucket interval, if it's giant, do we run the risk of not enqueueing and flushing for a while?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+	mgr.flushTimes.Push(newFlushMetadata)
+}
+
+func (mgr *leaderFlushManager) prepareFlushTimesWithLock(
+	buckets []*flushBucket,
+) schema.ShardSetFlushTimes {
+	proto := schema.ShardSetFlushTimes{
+		ByShard: make(map[uint32]*schema.ShardFlushTimes, defaultInitialFlushTimesCapacity),
+	}
+	for _, bucket := range buckets {
```
martin-mao: Can we parallelize the bucket preparations since this is done with the lock?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+	timeNanos int64
+	bucketIdx int
+}
+
+// flushMetadataHeap is a min heap for flush metadata where the metadata with the
+// earliest flush time will be at the top of the heap. Unlike the generic heap in
+// the container/heap package, pushing data to or popping data off of the heap doesn't
+// require conversion between flush metadata and interface{}, therefore avoiding the
+// memory and GC overhead due to the additional allocations.
+type flushMetadataHeap []flushMetadata
```
martin-mao: What's the max size of the heap? Is this one of those cases like in M3DB where we spend time to order a heap size of 3?

aggregator/list.go
```diff
@@ -42,31 +43,47 @@ var (
...
-		flushSuccess:   flushScope.Counter("success"),
-		flushErrors:    flushScope.Counter("errors"),
-		flushDuration:  flushScope.Timer("duration"),
+		encodeErrors:        scope.Counter("encode-errors"),
+		flushCollected:      flushScope.Counter("collected"),
+		flushSuccess:        flushScope.Counter("success"),
+		flushErrors:         flushScope.Counter("errors"),
+		flushDuration:       flushScope.Timer("duration"),
+		flushStale:          flushScope.Counter("stale"),
+		flushBeforeDuration: scope.Timer("flush-before.duration"),
```
martin-mao: Nit: Maybe rename this for pre-flush.duration?

aggregator/leader_flush_mgr.go
xichen2020: I'd rather keep it simple given that there's nothing that's time-consuming in this operation, which probably takes on the order of microseconds and is not the bottleneck. Also the lock is not heavily contended (the write lock is only needed when a new list is registered, which only happens when the server sees a new policy, which happens relatively rarely).

aggregator/leader_flush_mgr.go
xichen2020: Max size is the number of shards * number of resolutions per shard, which in the worst case is 15 * 6 = 90.

aggregator/list.go
xichen2020: Sure sounds good.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,505 @@
...
+func newElectionManagerMetrics(scope tally.Scope) electionManagerMetrics {
+	return electionManagerMetrics{
+		campaignCreateErrors: scope.Counter("campaign-create-errors"),
+		campaignErrors:       scope.Counter("campaign-errors"),
+		leaderErrors:         scope.Counter("leader-errors"),
+		leaderNotChanged:     scope.Counter("leader-not-changed"),
+		resignErrors:         scope.Counter("resign-errors"),
+		resignNotFollower:    scope.Counter("resign-not-follower"),
+		followerToLeader:     scope.Counter("follower-to-leader"),
+		leaderToFollower:     scope.Counter("leader-to-follower"),
```
xichen2020: These two are emitted when a follower to leader or a leader to follower transition happens, which are events and better captured as counters. There is a separate gauge `electionState` below for the purpose of determining what the current election state is and can be used to know who the leader is.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,505 @@
...
+}
+
+func (mgr *electionManager) Resign(ctx context.Context) error {
+	mgr.Lock()
+	if mgr.state != electionManagerOpen {
+		mgr.Unlock()
+		return errElectionManagerExpectOpen
+	}
+	if mgr.electionState == FollowerState {
+		mgr.Unlock()
```
xichen2020: It's not necessarily an error. Rather I think of it as a no op. Can emit a metric here sure.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,505 @@
...
+		return nil
+	}
+	mgr.state = electionManagerResigning
+	mgr.resignWg = &sync.WaitGroup{}
+	resignWg := mgr.resignWg
+	resignWg.Add(1)
```
xichen2020: The increment needs to be done inside the lock, otherwise if the election goroutine does a `resignWg.Wait()` after `mgr.Unlock()` but before `resignWg.Add(1)`, it may proceed prematurely without waiting for the necessary resignation operations to finish.

aggregator/election_mgr.go
xichen2020: Sure can do.

robskillington:
Still going - will post comments soon, although likely in the morning.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+		leaderService:   opts.LeaderService(),
+		leaderValue:     campaignOpts.LeaderValue(),
+		state:           electionManagerNotOpen,
+		doneCh:          make(chan struct{}),
+		notifyCh:        make(chan struct{}, 1),
+		electionState:   FollowerState,
+		sleepFn:         time.Sleep,
+		metrics:         newElectionManagerMetrics(instrumentOpts.MetricsScope()),
+	}
+	go mgr.reportMetrics()
```
robskillington: Shouldn't perhaps this be launched from `Open(...)`?  It seems the only way to break from the metrics loop is to read the close of the `doneCh` which can only be achieved if you've opened the manager.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+	mgr.state = electionManagerResigning
+	mgr.resignWg = &sync.WaitGroup{}
+	resignWg := mgr.resignWg
+	resignWg.Add(1)
+	mgr.Unlock()
+
+	err := mgr.leaderService.Resign(mgr.electionKey)
+	if err != nil {
+		resignWg.Done()
```
robskillington: Isn't it best to do `resignWg.Done()` this inside the lock considering others might wake up and acquire a read lock first and see that it's still resigning?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+	if isFollower {
+		return nil
+	}
+
+	// If the election state is not follower, it means we failed to establish
+	// the follower state with kv.
+	err = fmt.Errorf("instance resigned but is not follower: %v", ctx.Err())
+	mgr.metrics.resignNotFollower.Inc(1)
+	mgr.logError("resign failed", err)
+	return err
```
robskillington: At this point will the process still believe its a leader?  Seems strange and wondering what the correct course of action the caller can make at this point.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+		}
+		if ctxDone || isFollower {
+			break
+		}
+	}
+
+	// If the context expires, we cancel the in progress change and re-check
+	// the election state.
+	if !isFollower {
+		mgr.cancelInProgressChange()
```
robskillington: As far as I can tell there shouldn't be any change in progress right now?  We never assigned a cancel fn during this method.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+
+			// NB(xichen): campaign retrier retries forever until either the Campaign call succeeds,
+			// or the election manager has resigned.
+			if err := mgr.campaignRetrier.AttemptWhile(notDone, func() error {
+				var err error
+				campaignStatusCh, err = mgr.leaderService.Campaign(mgr.electionKey, mgr.campaignOpts)
+				if err == nil {
+					return nil
+				}
+				mgr.metrics.campaignCreateErrors.Inc(1)
```
robskillington: There should probably be at least 10-100ms of backoff here perhaps to avoid slamming downstream in case of fast failures?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+	if campaignStatus.State == campaign.Error {
+		mgr.metrics.campaignErrors.Inc(1)
+		mgr.logError("error campaigning", campaignStatus.Err)
+		return
+	}
+
+	currState := mgr.ElectionState()
+	newState := FollowerState
+	if campaignStatus.State == campaign.Leader {
+		newState = LeaderState
```
robskillington: Any reason not to have a mapping function here?  i.e. campaign.State -> ElectionStatus that returns an error if no exact match?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+func (mgr *electionManager) changeElectionState(currState, newState ElectionState) {
+	if currState == newState {
+		if newState != LeaderState {
+			return
+		}
+		// We are in the middle of verifying leader and changing from leader to follower
+		// and we received an update that indicates the current instance is now the leader,
+		// therefore we cancel the change from leader to follower and instead set the
+		// election state to leader.
+		if mgr.cancelInProgressChange() {
```
robskillington: Again I don't see any in progress change that happens during a resign?  i.e. cancel fn never gets set.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+
+	currState := mgr.ElectionState()
+	newState := FollowerState
+	if campaignStatus.State == campaign.Leader {
+		newState = LeaderState
+	}
+	mgr.changeElectionState(currState, newState)
+}
+
+func (mgr *electionManager) changeElectionState(currState, newState ElectionState) {
```
robskillington: Shouldn't currState be read while in a lock from the struct and all this logic be done while holding the lock with the read current state of the world?  Otherwise there is racy conditions when preconditions are slightly different going into this method at about the same time.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+	if currState == newState {
+		if newState != LeaderState {
+			return
+		}
+		// We are in the middle of verifying leader and changing from leader to follower
```
robskillington: It's strange that this change status would be called and both currState and newState is leader no?  

Also isn't currState and newState Leader?  Why have we determined we're trying to become a follower if both currState and newState is LeaderState?

Maybe there needs to be a third state like `pendingFollower` if that's the current state you're trying to capture here?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+		case <-ctx.Done():
+			return false
+		default:
+			return true
+		}
+	}
+
+	// NB(xichen): majority of the verification logic below will be moved into the
+	// leadership API implementation to ensure the API user will receive a follower
+	// status only after this verification is performed.
```
robskillington: Any reason not to do that move now?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,518 @@
...
+	mgr.changeRetrier.AttemptWhile(continueFn, func() error {
+		leader, err := mgr.leaderService.Leader(mgr.electionKey)
+		if err != nil {
+			mgr.metrics.leaderErrors.Inc(1)
+			mgr.logError("error determining the leader", err)
+			return err
+		}
+		if leader == mgr.leaderValue {
```
robskillington: Isn't there some read lock that should be acquired to test against `mgr.leaderValue`?

aggregator/aggregator.go
```diff
@@ -192,14 +221,30 @@ func (agg *aggregator) Open() error {
 	defer agg.Unlock()
 
 	if agg.state != aggregatorNotOpen {
-		return errAggregatorIsOpenOrClosed
+		return errAggregatorAlreadyOpenOrClosed
```
martin-mao: What's NotOpen, like it hasn't been opened ever yet?

aggregator/follower_flush_mgr.go
martin-mao: You can record each group inside their go routine no?

aggregator/follower_flush_mgr.go
martin-mao: But we won't have a case where for some reason our flush time is like 100ms (or 1s) and we just can't flush everything in time, causing more and more go routines to be blocked on this pool eventually?

aggregator/aggregator.go
xichen2020: `NotOpen` is the uninitialized state, which means the aggregator has never been opened before. Right now the semantics is that if the aggregator has been closed (which usually happens during a process shutdown), it cannot be re-opened. That semantics is hard to achieve with only two states.

aggregator/follower_flush_mgr.go
xichen2020: I could but the flushes are then done sequentially within the group even though the groups are processed in parallel. Compared to flushing the groups sequentially but processing the individual flushes in parallel, the two approaches will probably take the same amount of time in the end.

aggregator/follower_flush_mgr.go
xichen2020: That won't happen because the flushes are performed sequentially, i.e., the next flush won't even start until the current flush finishes, so the goroutines will always be returned back to pool before starting the next flush.

aggregator/aggregator.go
martin-mao: For sure, maybe calling it aggregatorUninitialized would be better? Bit confusing between NotOpen and Closed.

aggregator/follower_flush_mgr.go
martin-mao: What's the behavior if the flush takes longer than the smallest flush interval? Will you try to flush asap once this loop is done or do you miss a whole flush interval?

aggregator/election_mgr.go
xichen2020: Yes the process still assumes the leader role. This corresponds to the scenario where the process has resigned but before it can confirm the leader role has been successfully transferred to another instance, it loses connectivity with etcd (e.g. network partitioning, etcd cluster going down etc). If that happens, the process should behave conservatively and continue assuming the leader role and flushing metrics downstream until it can reestablish connection with the etcd cluster. This is better than the alternative where the instance prematurely assumes a follower role and stops flushing, even though the leader role has not been successfully transferred (e.g., the other node died, or the etcd cluster died before the other node could take over the leader role).

Once it reconnects with the etcd cluster, it'll be able to restart the campaign and determine its election state properly.

aggregator/election_mgr.go
xichen2020: Yes this is what `campaignRetrier` is for. I currently use an initial backoff of 100ms and a backoff factor of 2 with a max backoff of 2s for the retrier.

aggregator/election_mgr.go
xichen2020: Sure can add a mapping function.

aggregator/election_mgr.go
xichen2020: Yeah fair point, will address.

aggregator/election_mgr.go
xichen2020: Yeah so the scenario I'm trying to capture here is that there is an ongoing leader to follower transition as a result of a previous election status change, and now we have received a new update that says we are now the leader, so we should cancel the ongoing leader to follower transition.

Probably makes sense to add something like `pendingFollower`, will update.

aggregator/election_mgr.go
xichen2020: So during a resign, `leaderService.Resign()` is called, and if the current instance is a leader, the leader election library (in `m3cluster/services/leader` package by this PR) forces the instance to resign from the current leader role and pushes a follower status to the `campaignStatusCh`, which gets picked up by the election goroutine on line 354 in `startElectionLoop()` and the status update is processed in `changeElectionState()`.

aggregator/election_mgr.go
xichen2020: @schallert will work on this as a separate PR, tracked in https://github.com/m3db/m3cluster/issues/133.

generated/proto/flush/flush.proto
```diff
@@ -0,0 +1,28 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+syntax = "proto3";
+
+message ShardSetFlushTimes {
+  map<uint32, ShardFlushTimes> by_shard = 1;
```
cw9: is the uint32 for ShardSetID? it's defined as string in m3cluster right now, should we make them the same type?

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,350 @@
...
+	mgr.RLock()
+	defer mgr.RUnlock()
+
+	if mgr.flushTimesState != flushTimesProcessed {
+		mgr.metrics.flushTimesNotProcessed.Inc(1)
+		return false
+	}
+
+	allWindowsEnded := true
+	for _, shardFlushTimes := range mgr.proto.ByShard {
```
cw9: Is this safe around a topology change? For example, when another shardSet was removed from the placement, this shardset may get assigned a few more shards, what if the follower can lead the old shards but not the new ones?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+func (mgr *electionManager) processGoalState(goalState goalState) {
+	mgr.Lock()
+	defer mgr.Unlock()
+
+	currState := mgr.electionState
+	newState := goalState.state
+	if currState == newState {
+		return
+	}
+	if currState == FollowerState && newState == PendingFollowerState {
```
robskillington: Hm is this a valid transition?  If not maybe we should log.Warnf here?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+			if leader == mgr.leaderValue {
+				isLeader = true
+				mgr.metrics.verifyLeaderNotChanged.Inc(1)
+				mgr.logError("leader has not changed", errLeaderNotChanged)
+				return errLeaderNotChanged
+			}
+			isLeader = false
+			return nil
+		})
+		cancel()
```
robskillington: Do you need to bother calling cancel considering no one else now needs to depend on this context?  i.e. it's only used for timeout in the `continueFn`?

aggregator/election_mgr.go
xichen2020: Conceptually it isn't, but because the leader election API currently does not verify states, if the client receives a follower state, it can only treat it as a "pendingFollower" state. Once the leader election logic is adapted to only notify clients of states that have been verified (@mschalle and I just talked offline and he's working on it), this is no longer needed. I'll put a comment here making a note of it.

aggregator/election_mgr.go
xichen2020: Apparently cancel is used to release resources associated with contexts in addition to cancelling them. If you don't call `cancel` explicitly, intellisense complains.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+	mgr.goalStateLock.Lock()
+	mgr.setGoalStateWithLock(newState)
+	mgr.goalStateLock.Unlock()
+}
+
+func (mgr *electionManager) setGoalStateWithLock(newState ElectionState) {
+	newGoalStateID := mgr.goalStateID
+	mgr.goalStateID++
+	newGoalState := goalState{id: newGoalStateID, state: newState}
+	mgr.goalStateWatchable.Update(newGoalState)
```
robskillington: Hm shouldn't this also update the `mgr.electionState`?

xichen2020: This method only sets the goal state, `processGoalState()` is the method that actually watches the goal state watchable and updates the election state.

aggregator/election_mgr.go
robskillington: Sounds good.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+
+	if mgr.state != electionManagerOpen {
+		return errElectionManagerExpectOpen
+	}
+	close(mgr.doneCh)
+	mgr.state = electionManagerClosed
+	return nil
+}
+
+func (mgr *electionManager) watchStateChanges(watch xwatch.Watch) {
```
robskillington: Can we call this `watchGoalStateChanges`?  Just so its clear that we're not watching state updates from server in this loop?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+				continue
+			}
+		case <-mgr.doneCh:
+			return
+		}
+		mgr.processStatusUpdate(campaignStatus)
+	}
+}
+
+func (mgr *electionManager) processStatusUpdate(campaignStatus campaign.Status) {
```
robskillington: Can we rename this `processCampaignUpdate(...)`?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+		return
+	}
+	mgr.RUnlock()
+
+	if err := mgr.leaderService.Resign(mgr.electionKey); err != nil {
+		action.errCh <- err
+	}
+}
+
+func (mgr *electionManager) startElectionLoop() {
```
robskillington: Can we rename this `campaignLoop()`?

aggregator/election_mgr.go
xichen2020: Sure thing.

aggregator/election_mgr.go
xichen2020: Sure thing.

aggregator/election_mgr.go
xichen2020: Sure thing.

aggregator/election_mgr.go
robskillington: Ouch, ok then.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,544 @@
...
+			return nil
+		})
+		cancel()
+
+		mgr.goalStateLock.Lock()
+		state := mgr.goalStateWatchable.Get()
+		// If the latest goal state is different from the goal state before verification, this
+		// means the goal state has been updated since verification started and the pending follower
+		// to follower transition is no longer valid.
+		if state != currState {
```
robskillington: As discussed perhaps explicitly check the ID here?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+			lastFlushedNanos := flusher.LastFlushedNanos()
+			shardFlushTimes.ByResolution[int64(resolution)] = lastFlushedNanos
+		}
+	}
+	return proto
+}
+
+func (mgr *leaderFlushManager) persistFlushTimes(flushTimes *schema.ShardSetFlushTimes) {
+	persistStart := mgr.nowFn()
+	_, err := mgr.flushTimesStore.Set(mgr.flushTimesKey, flushTimes)
```
robskillington: Should probably retry on error yeah?  Otherwise won't follower need to flush these times?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+	flushTimes    *schema.ShardSetFlushTimes
+}
+
+func (t *leaderFlushTask) Run() {
+	mgr := t.mgr
+
+	// NB(xichen): we intentionally do not use the worker pool as persistence could be
+	// done asynchronously to avoid blocking the workers reserved for flushing.
+	if t.shouldPersist {
+		go mgr.persistFlushTimes(t.flushTimes)
```
robskillington: Shouldn't this only be done on success?

aggregator/leader_flush_mgr.go
robskillington: Yeah shouldn't we do a jitter max per bucket?  Otherwise if we have 1 hour tiles it could be a full 59 minutes after the the hour bucket is aggregated before we flush.

Even for a 10 minute bucket we could be 9 minutes late.  I should think we want this to be configurable per bucket.  Or at the very least some fixed % of the `flushInterval` size (say 10% - but even then this would likely want to be configurable per bucket size as you're probably ok for 1 second buckets to be 100% of the flush interval to be jittered).

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+			// NB(xichen): don't oversleep if the next flush is about to happen.
+			timeToNextFlush := time.Duration(earliestFlush.timeNanos - nowNanos)
+			if timeToNextFlush < waitFor {
+				waitFor = timeToNextFlush
+			}
+		}
+	}
+
+	var (
+		now           = mgr.nowFn()
```
robskillington: This seems to be what you're saying "we are flushed up to" - shouldn't this be `nowNanos` from above, as this is the actual value you compare against buckets to?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+			bucketIdx := earliestFlush.bucketIdx
+			duration = buckets[bucketIdx].duration
+			flushers = buckets[bucketIdx].flushers
+			nextFlushMetadata := flushMetadata{
+				timeNanos: earliestFlush.timeNanos + int64(buckets[bucketIdx].interval),
+				bucketIdx: bucketIdx,
+			}
+			mgr.flushTimes.Pop()
+			mgr.flushTimes.Push(nextFlushMetadata)
+			mgr.flushedSincePersist = true
```
robskillington: Shouldn't `waitFor` be set to 0 in this code path?  Otherwise how do you ever catch up if you're more than one bucket behind?  

For instance if you ever have more than one bucket waiting to be flushed you'll enter this `nowNanos >= earliestFlush.timeNanos` and do a single flush task but return `waitFor` to be the the `checkEvery` value.  This means you'll never clear out your backlog.

xichen2020: Yeah so this is handled below at line 170 where if we need to do a flush, we always return 0 as the `waitFor` value, but I can set it explicitly here sure thing.

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+func (mgr *leaderFlushManager) Prepare(buckets []*flushBucket) (flushTask, time.Duration) {
+	var (
+		shouldFlush = false
+		duration    tally.Timer
+		flushers    []PeriodicFlusher
+		waitFor     = mgr.checkEvery
+	)
+	mgr.Lock()
```
robskillington: Any reason not to `defer mgr.Unlock()` here to avoid a bug being introduced in the future?  Seems we only save on a few extra CPU instructions at the bottom setting values by not using defer here.

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,350 @@
...
+func (mgr *followerFlushManager) Prepare(buckets []*flushBucket) (flushTask, time.Duration) {
+	var (
+		shouldFlush        = false
+		flushersByInterval []flushersGroup
+	)
+	// NB(xichen): a flush is triggered in the following scenarios:
+	// * The flush times persisted in kv have been updated since last flush.
+	// * Sufficient time (a.k.a. maxNoFlushDuration) has elapsed since last flush.
+	now := mgr.nowFn()
+	mgr.Lock()
```
robskillington: Any reason not to `defer mgr.Unlock()` here to avoid a bug being introduced in the future?  Seems we only save on a few extra CPU instructions at the bottom setting values by not using defer here.

aggregator/leader_flush_mgr.go
xichen2020: Sure yeah, can use `nowNanos`.

aggregator/leader_flush_mgr.go
xichen2020: Sure yeah can make it per flush bucket.

aggregator/leader_flush_mgr.go
xichen2020: So if it fails to persist, the follower will keep the metrics around for a bit longer (till when persistence succeeds next time), which is probably ok, but doesn't hurt to add a retrier, sure can do that.

aggregator/leader_flush_mgr.go
xichen2020: Not sure what you meant by "only be done on success". On success of ?

aggregator/flush_mgr_options.go
```diff
@@ -35,7 +37,11 @@ const (
...
+	defaultFlushTimesKeyFormat    = "/shardset/%s/flush"
+	defaultFlushTimesPersistEvery = 10 * time.Second
+	defaultMaxNoFlushDuration     = 15 * time.Minute
```
robskillington: Maybe this should be 5 minutes by default?  Or even 2 minutes.  That's pretty far behind for all metrics to be behind by if leader goes down no?  Or would it auto-elect itself before that time?

aggregator/leader_flush_mgr.go
xichen2020: Sure can use `defer`.

aggregator/follower_flush_mgr.go
xichen2020: Sure can use `defer`.

aggregator/flush_mgr_options.go
xichen2020: `maxNoFlushDuration` is how long the follower keeps metrics in memory if it fails to observe an flush timestamp update from kv. If the leader goes down, the follower will automatically be promoted to follower so that doesn't apply here. This is an extra measure put in place in case the leader is up but not persisting flush times (e.g., etcd is down, or a bug in the code), in which case the follower will hold on to the metrics in the meantime until either it receives a kv update, or 15 minutes has passed with no kv update, in which case the follower will flush the oldest 10s of data. Ideally this value should be derived from the % of memory reserved for holding metrics in memory for this particular scenario. Additionally, alerts should be set up if the duration without flushes exceeds e.g., 30s if the leader is supposed to persist the times to kv every 10s.

aggregator/follower_flush_mgr.go
```diff
@@ -0,0 +1,350 @@
...
+	for _, shardFlushTimes := range mgr.proto.ByShard {
+		for windowNanos, lastFlushedNanos := range shardFlushTimes.ByResolution {
+			windowSize := time.Duration(windowNanos)
+			windowEndAt := mgr.openedAt.Truncate(windowSize)
+			if windowEndAt.Before(mgr.openedAt) {
+				windowEndAt = windowEndAt.Add(windowSize)
+			}
+			if lastFlushedNanos < windowEndAt.UnixNano() {
+				mgr.metrics.flushWindowsNotEnded.Inc(1)
+				allWindowsEnded = false
```
robskillington: Can't you just return false here instead of all the breaking, etc?

services/m3aggregator/main/main.go
```diff
@@ -121,6 +121,7 @@ func main() {
 
 	select {
 	case <-closedCh:
+		aggregator.Close()
```
robskillington: Shouldn't this happen in the goroutine just before `close(closedCh)`?  Otherwise it could block and it my surpass the gracefulShutdownTimeout

aggregator/flush_mgr_options.go
robskillington: I see, no problems - sounds good.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+	errElectionManagerAlreadyOpenOrClosed = errors.New("election manager is already open or closed")
+	errElectionManagerNotOpenOrClosed     = errors.New("election manager is not open or closed")
+	errLeaderNotChanged                   = errors.New("leader has not changed")
+	errResignActionAlreadyEnqueued        = errors.New("resign action already enqueued")
+)
+
+type electionManagerState int
+
+const (
+	electionManagerNotOpen electionManagerState = iota
```
martin-mao: Nit: would electionManagerUninitalized be a better name given Open and Closed seem to be the 2 states something can be in.

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+	mgr.state = electionManagerOpen
+	mgr.electionKey = fmt.Sprintf(mgr.electionKeyFmt, shardSetID)
+
+	_, stateChangeWatch, err := mgr.goalStateWatchable.Watch()
+	if err != nil {
+		return err
+	}
+	go mgr.watchGoalStateChanges(stateChangeWatch)
+
+	_, verifyWatch, err := mgr.goalStateWatchable.Watch()
```
martin-mao: Nit: If this fails, should we not execute line 259? ie. move that line to below this check?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+	if currState == newState {
+		return
+	}
+	// NB(xichen): the follower to pending follower transition is conceptually invalid
+	// but because the leader election API currently does not verify states, if the client
+	// receives a follower state, it can only treat it as a "pendingFollower" state. This
+	// is no longer needed once the leader election logic is adapted to only notify clients
+	// of states that have been verified.
+	if currState == FollowerState && newState == PendingFollowerState {
+		return
```
martin-mao: Worth logging even if it's impossible to happen?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+			return
+		case <-watch.C():
+		}
+
+		currState := mgr.goalStateWatchable.Get().(goalState)
+		if currState.state != PendingFollowerState {
+			continue
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), mgr.changeVerifyTimeout)
```
martin-mao: Do we not need to close the context at some point?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+func (mgr *electionManager) resign(action resignAction) {
+	mgr.RLock()
+	if mgr.state != electionManagerOpen {
+		action.errCh <- errElectionManagerNotOpenOrClosed
+		mgr.RUnlock()
+		return
+	}
+	mgr.RUnlock()
+
+	if mgr.ElectionState() == FollowerState {
```
martin-mao: Is it safe to read ElectionState outside of lock?

aggregator/election_mgr.go
```diff
@@ -0,0 +1,545 @@
...
+		return
+	}
+
+	mgr.goalStateLock.Lock()
+	mgr.setGoalStateWithLock(newState)
+	mgr.goalStateLock.Unlock()
+}
+
+func (mgr *electionManager) setGoalStateWithLock(newState ElectionState) {
+	newGoalStateID := mgr.goalStateID
```
martin-mao: Nit: Is it weird calling this newGoalStateID since we increment it right after, maybe just goalStateID? Also, is it weird that mgr.goalStateID is actually 1 more than the current goalStateID?

aggregator/election_mgr.go
xichen2020: So `NotOpen`, `Open` and `Closed` are the three states we've consistently used in all of our repos. Technically speaking `NotOpen` is also a state. I'd prefer to stick to the convention, but can change if you feel strongly about it.

aggregator/election_mgr.go
xichen2020: Sure can do, this is called on aggregator startup, so if this fails, the process will exit, but doesn't hurt to be defensive.

aggregator/election_mgr.go
xichen2020: As stated in the comments, right now it may still happen due to the leader election API not verifying states returned to the caller, I can potentially add a metric here, will do that.

aggregator/election_mgr.go
xichen2020: We do, although this is a golang context and the standard way of closing it is to call `cancel` down below at line 402.

aggregator/election_mgr.go
xichen2020: Yes it's safe, `mgr.ElectionState()` does a `Get` on the underlying `electionStateWatchable`, which is thread-safe.

aggregator/election_mgr.go
xichen2020: Sure, can change `goalStateID` to `nextGoalStateID` if that reads better.

martin-mao:
LGTM

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,371 @@
...
+func (t *leaderFlushTask) Run() {
+	mgr := t.mgr
+
+	// NB(xichen): we intentionally do not use the worker pool as persistence could be
+	// done asynchronously to avoid blocking the workers reserved for flushing. In theory
+	// if the persistence takes a long to finish, there is a chance two goroutines may
+	// persist flush times at the same time. However in practice this is okay because kv
+	// persistence is fast so this rarely if at all happens. And even if it happens, the
+	// impact is that the follower may flush less data than it optimally could but that
+	// does not affect the correctness of the system.
```
cw9: nit: why not just order the kv.Set() by having a thread listen on a channel for notification, and in Run() just send a signal to the channel? I know the current logic is not incorrect. But just to be more careful here?

aggregator/leader_flush_mgr.go
```diff
@@ -0,0 +1,371 @@
...
+	if t.shouldPersist {
```
cw9: should this happen after the `shouldFlush{}` block?

aggregator/list.go
```diff
@@ -196,15 +241,17 @@ func (l *metricList) Flush() {
...
-		newEncoder := l.encoderPool.Get()
-		newEncoder.Reset()
-		l.encoder.Reset(newEncoder)
-		if err := l.flushHandler.Handle(NewRefCountedBuffer(encoder)); err != nil {
-			l.log.Errorf("flushing metrics error: %v", err)
-			l.metrics.flushErrors.Inc(1)
-		} else {
-			l.metrics.flushSuccess.Inc(1)
+	if flushType == consumeType {
+		if encoder := l.encoder.Encoder(); len(encoder.Bytes()) > 0 {
```
cw9: nit: `if encoder := l.encoder.Encoder(); len(encoder.Bytes()) > 0 && flushType == consumeType {...}` to save an indentation?

services/m3aggregator/config/aggregator.go
```diff
@@ -515,32 +547,210 @@ func (t shardFnType) ShardFn() (aggregator.ShardFn, error) {
...
+	if c.ResignTimeout != 0 {
+		opts = opts.SetResignTimeout(c.ResignTimeout)
+	}
+	if c.TTLSeconds != 0 {
+		opts = opts.SetTTLSecs(c.TTLSeconds)
+	}
+	return opts, nil
+}
+
+type serviceIDConfiguration struct {
```
cw9: nit: could you add a TODO: move this to m3cluster and reuse in etcd/config.go?

PR #38, opened 2017-08-02
Add deployment logic for deploying the aggregation servers

xichen2020:
cc @prateek @martin-mao @jeromefroe @cw9 

This PR adds deployment logic to deploy the aggregation servers. The main file and the config file will live internally due to the deployment manager using uber-specific deployment system APIs.


tools/deploy/client.go
```diff
@@ -0,0 +1,120 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/m3db/m3aggregator/aggregator"
+	httpserver "github.com/m3db/m3aggregator/server/http"
+)
+
+// AggregatorClient interacts with aggregator instances via aggregator HTTP endpoints.
+type AggregatorClient interface {
+	// IsHealthy determins whether an instance is healthy.
```
jeromefroe: nit: s/determins/determines

tools/deploy/helper.go
```diff
@@ -0,0 +1,370 @@
...
+		if di.Revision() == revision {
+			continue
+		}
+		filtered = append(filtered, placementInstances[i])
+	}
+	return filtered
+}
+
+// validateInstances validates instances derived from placement against
+// instaces derived from deployment, ensuring there are no duplicate instances
```
jeromefroe: nit: s/instaces/instances

tools/deploy/client.go
```diff
@@ -0,0 +1,120 @@
...
+	IsHealthy(instance string) error
+
+	// Status returns the instance status.
+	Status(instance string) (aggregator.RuntimeStatus, error)
+
+	// Resign resigns an aggregator instance.
```
prateek: nit: could you describe the semantics of the resign here - master gives up 'mastership' and doesn't contest election for ...

tools/deploy/helper.go
```diff
@@ -0,0 +1,370 @@
...
+		client:             opts.AggregatorClient(),
+		store:              opts.KVStore(),
+		retrier:            opts.Retrier(),
+		workers:            opts.WorkerPool(),
+		placementWatcher:   placementWatcher,
+		settleBetweenSteps: opts.SettleDurationBetweenSteps(),
+	}, nil
+}
+
+func (h deploymentHelper) Deploy(revision string) error {
```
prateek: Could you validate the `revision` before doing any of the subsequent operations?

tools/deploy/helper.go
```diff
@@ -0,0 +1,370 @@
...
+	"github.com/m3db/m3x/retry"
+	"github.com/m3db/m3x/sync"
+)
+
+var (
+	errNoDeploymentProgress = errors.New("no deployment progress")
+)
+
+// DeploymentHelper is helper class handling deployments.
+type DeploymentHelper interface {
```
prateek: nit: rename to `Helper` instead, so users would use `deploy.Helper` instead of `deploy.DeploymentHelper`

tools/deploy/helper.go
```diff
@@ -0,0 +1,370 @@
...
+	// Deploy deploys a target revision to the instances in the placement.
+	Deploy(revision string) error
+}
+
+// TODO(xichen): disable deployment while another is ongoing.
```
prateek: still applicable for this pr?

tools/deploy/helper.go
```diff
@@ -0,0 +1,370 @@
...
+	client             AggregatorClient
+	store              kv.Store
+	retrier            xretry.Retrier
+	workers            xsync.WorkerPool
+	placementWatcher   services.StagedPlacementWatcher
+	settleBetweenSteps time.Duration
+}
+
+// NewDeploymentHelper creates a new deployment helper.
+func NewDeploymentHelper(opts DeploymentHelperOptions) (DeploymentHelper, error) {
```
prateek: nit: rename to `NewHelper`

tools/deploy/validate.go
```diff
@@ -0,0 +1,142 @@
...
+type factory struct {
+	client  AggregatorClient
+	workers xsync.WorkerPool
+}
+
+func newValidatorFactory(client AggregatorClient, workers xsync.WorkerPool) validatorFactory {
+	return factory{client: client, workers: workers}
+}
+
+func (f factory) ValidatorFor(
```
prateek: Does this need to be an exported function?

tools/deploy/client.go
```diff
@@ -0,0 +1,120 @@
...
+	Resign(instance string) error
+}
+
+type client struct {
+	httpClient *http.Client
+}
+
+// NewClient creates a new client.
+func NewClient(httpClient *http.Client) AggregatorClient {
+	return &client{httpClient: httpClient}
```
prateek: Could you do a sanity check and ensure the client has a reasonable timeout set

tools/deploy/helper_options.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/m3db/m3x/sync"
+)
+
+const (
+	defaultSettleDurationBetweenSteps = time.Minute
+	defaultHelperWorkerPoolSize       = 16
+)
+
+// DeploymentHelperOptions provide a set of options for the deployment helper.
+type DeploymentHelperOptions interface {
```
prateek: per above, rename to `HelperOptions`

tools/deploy/manager.go
```diff
@@ -0,0 +1,48 @@
...
+
+import "github.com/m3db/m3cluster/services"
+
+// DeploymentInstance represents an instance in the deployment.
+type DeploymentInstance interface {
+	// ID returns the instance id.
+	ID() string
+
+	// Revision returns the revision deployed to the instance.
+	Revision() string
```
prateek: Maybe make revision a type alias to a string instead.

tools/deploy/validate.go
```diff
@@ -0,0 +1,142 @@
...
+	"sync"
+
+	"github.com/m3db/m3aggregator/aggregator"
+	"github.com/m3db/m3cluster/services"
+	"github.com/m3db/m3x/sync"
+)
+
+// Validator performs validation on a deployment target, returning nil if
+// validation succeeds, and an error otherwise.
+type Validator func() error
```
prateek: nit (take it or leave it): rename as `ValidateFn` to make it clear it's a fn type

tools/deploy/helper.go
xichen2020: Not for this PR, but thankfully not too big an issue since the deployment tool will only deploy when nothing is actively deployed.

tools/deploy/helper.go
xichen2020: Sure thing.

tools/deploy/helper.go
xichen2020: Sure I can check whether it's empty or not, don't want to put too many constraints on what `revision` could be or otherwise it's going to compromise the genericity of the deployment tool.

tools/deploy/helper.go
xichen2020: Yeah good catch, ta.

tools/deploy/helper_options.go
xichen2020: Sure thing.

tools/deploy/manager.go
xichen2020: That's probably insufficient if we want to make the revision type truly generic, I'd actually make it an interface with `Validate` and `Equals` so users can provide their own implementation. Otherwise we'd end up doing string comparisons in the code anyway.

I think using string for revision is probably okay, and if the need for a more generic revision type arises in the future, can always easily change it.

tools/deploy/validate.go
xichen2020: That's what I started with but `ValidateFnFactory` is not as good as `ValidatorFactory` on the eyes.

tools/deploy/validate.go
xichen2020: Also, the interface itself is not exported so you can't actually use it outside the package. I tend to use public APIs in interfaces because if we ever want to export the interface in the future, you only need to change the interface name without touching the method names.

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+			continue
+		}
+		filtered = append(filtered, metadata)
+	}
+	return filtered
+}
+
+// instanceMetadata contains instance metadata.
+type instanceMetadata struct {
+	PlacementID  string
```
prateek: nit: rename to `instanceID` or `placementInstanceID` instead of `PlacementID` to avoid ambiguity with `service.Placement`

tools/deploy/planner.go
```diff
@@ -0,0 +1,263 @@
...
+	for shardSetID, group := range grouped {
+		shardSetID, group := shardSetID, group
+		wg.Add(1)
+		p.workers.Go(func() {
+			defer wg.Done()
+
+			electionKey := fmt.Sprintf(p.electionKeyFmt, shardSetID)
+			leader, err := p.leaderService.Leader(electionKey)
+			if err != nil {
+				err = fmt.Errorf("unable to determine leader for shard set id %s", shardSetID)
```
prateek: a few nits: 
- these four lines are repeated here and after the loop below, refactor to share same function?
- this line is dropping the actual error returned by `p.leaderService.Leader(...)`, could you add it to the error message

tools/deploy/planner.go
```diff
@@ -0,0 +1,263 @@
...
+		if !exists {
+			continue
+		}
+		group.All = append(group.All, instance)
+	}
+
+	// Determine the leader of each group.
+	var (
+		wg    sync.WaitGroup
+		errCh = make(chan error, 1)
```
prateek: Feels weird to be dropping all but the first error like this. Could you add a concurrent multi-error implementation in `xerrors.MultiError` and re-use that here?

tools/deploy/helper_options.go
```diff
@@ -0,0 +1,254 @@
...
+// ToPlacementInstanceIDFn converts a deployment instance id to the corresponding
+// placement instance id.
+type ToPlacementInstanceIDFn func(deploymentInstanceID string) (string, error)
+
+// ToAPIEndpointFn converts a placement instance endpoint to the corresponding
+// aggregator instance api endpoint.
+type ToAPIEndpointFn func(placementEndpoint string) (string, error)
+
+// HelperOptions provide a set of options for the deployment helper.
+type HelperOptions interface {
```
prateek: Add a `Validate()` which ensures required fields w/o default values are set (e.g. Manager)

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+	filtered := filterByRevision(all, revision)
+	plan, err := h.planner.GeneratePlan(filtered, all)
+	if err != nil {
+		return fmt.Errorf("unable to generate deployment plan: %v", err)
+	}
+
+	// If in dry run mode, log the generated deployment plan and return.
+	if mode == DryRunMode {
+		h.logger.Infof("Generated deployment plan: %+v", plan)
```
prateek: probably useful to list the deployment plan even if not in dryRunMode.

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+	placement, onPlacementDoneFn, err := stagedPlacement.ActivePlacement()
+	if err != nil {
+		return nil, err
+	}
+	defer onPlacementDoneFn()
+
+	return placement.Instances(), nil
+}
+
+func filterByRevision(metadatas instanceMetadatas, revision string) instanceMetadatas {
```
prateek: nit: make this a method on `instanceMetadatas`,

suggestion (take/leave): rename to `withRevision`

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+	for i, step := range plan.Steps {
+		h.logger.Infof("deploying step %d of %d", i+1, numSteps)
+		if err := h.executeStep(step, revision, toDeploy, all); err != nil {
+			return err
+		}
+		if i < numSteps-1 && h.settleBetweenSteps > 0 {
+			h.logger.Infof("waiting settle duration after step: %s", h.settleBetweenSteps.String())
+			time.Sleep(h.settleBetweenSteps)
+		}
+		h.logger.Infof("deploying step %d succeeded", i+1)
```
prateek: nit: maybe re-order this log to be above the settle duration sleep case to give end users more information earlier

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+			return fmt.Errorf("only %d out of %d instances are safe to deploy", safe, len(instances))
+		}
+		return nil
+	})
+}
+
+func (h helper) validate(targets []deploymentTarget) error {
+	var (
+		wg    sync.WaitGroup
+		errCh = make(chan error, 1)
```
prateek: looks like you could re-use the concurrent multi error here too.

tools/deploy/helper.go
```diff
@@ -0,0 +1,442 @@
...
+func (h helper) Deploy(revision string, mode Mode) error {
+	if revision == "" {
+		return errInvalidRevision
+	}
+	all, err := h.allInstanceMetadatas()
+	if err != nil {
+		return fmt.Errorf("unable to get all instance metadatas: %v", err)
+	}
+	filtered := filterByRevision(all, revision)
+	plan, err := h.planner.GeneratePlan(filtered, all)
```
prateek: A general suggestion: It'd be more robust to treat leader->follower elections as common place occurrences, even if they're not going to happen frequently. For e.g. instead of generating the entire plan up front, it'd be better if we generated a step of the plan, executed it, then generated the next step, etc..

tools/deploy/helper_options.go
xichen2020: Sure can do

tools/deploy/planner.go
xichen2020: I think this is a pattern we do use in other places as well. The rationale here is we don't want to proceed as soon as an error happens and the errors are likely repetitive due to the same reason (e.g., underlying etcd issue). I could use a multi error here but given that there are many instances I worry that may pollute the logs too much when say the underlying etcd cluster goes down.

tools/deploy/planner.go
xichen2020: Re 1: not sure which four lines you are referring to?
Re 2: yeah good catch, will add back the error.

tools/deploy/helper.go
xichen2020: We kind already do but instead print out the whole plan, we print out the targets in each step, which arguable is more clear.

tools/deploy/helper.go
xichen2020: Sure yeah can do.

tools/deploy/helper.go
xichen2020: Sure can do.

tools/deploy/helper.go
xichen2020: `instanceID` is also confusing, thought about `placementInstanceID` but it was too verbose. `PlacementID` is probably okay given that it's inside an `instanceMetadata` object, which should scope the field to a single instance.

tools/deploy/helper.go
xichen2020: I tend to agree with that approach if the leader / follower transitions occur very frequently, in reality however they happen rarely (e.g., during deployments or single node failures), so if we generate the plan upfront and execute one step at a time, it'll succeed 99% of the time, and for the remaining 1% we can simply restart the deployment tool, which makes the logic flow simpler. This is also the deployment strategy m3admin uses.

tools/deploy/helper.go
prateek: I wasn't sure what placementID meant until I looked at how the struct was being created/used, especially cause the staged placement handler made me think there could be multiple placements per service instance. 

How about renaming to `id`, or adding a comment to indicate it's the placement instance's ID.

tools/deploy/helper.go
prateek: I was thinking that it might be useful to the operator to know what the next steps are before they're run. For instance to know when a particular group was going to be deployed.

Here's how I'd probably run this tool:
(1) start a screen/tmux session
(2) do the dry run and ensure stuff makes sense
(3) do the actual run and `tee` to a log. 
(4) break out of the screen/tmux session
(5) follow along the `tee`d log to see what's going on. 

I can imagine not remembering the output of the dry run once the execution has started, or not `tee`-ing it. At which point I'd have to re-run the dry run mode to see what the executor is going to do next (would that even generate the same output btw?). Would be easier to just log the plan up-front before the execution. 

Feel free to disregard if you don't think that's a valid use-case.

tools/deploy/planner.go
prateek: Re 1:
```
err = fmt.Errorf("unknown leader %s for shard set id %s", leader, shardSetID)
select {
  case errCh <- err:
  default:
}
```

tools/deploy/planner.go
prateek: The reason that feels weird is that we're waiting for all the go-routines to finish but only capturing the error from the first one to report an issue. If one error is sufficient for us to terminate execution, we should do that and end the function as soon as any of the go routines run into an error. If we're going to wait for all the errors, then we should capture them all. 

The MultiError suggestion was to try and and abstract the errorCh logic which seems to be replicated in a few functions here.

tools/deploy/helper.go
prateek: My main point is we're building tooling assuming that's not going to change when we don't need to. If we do figure out a way to live with frequent leader/follower transitions, it'd be nice not to have to re-write the tooling. Judgement call, so up to you.

tools/deploy/helper.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+// validateInstances validates instances derived from placement against
+// instances derived from deployment, ensuring there are no duplicate instances
+// and the instances derived from two sources match against each other.
+func (h helper) computeInstanceMetadatas(
+	placementInstances []services.PlacementInstance,
+	deploymentInstances []Instance,
+) (instanceMetadatas, error) {
+	if len(placementInstances) != len(deploymentInstances) {
```
cw9: Would we do deployments on hot spares in the future? In which case they will not be in the placement

tools/deploy/helper.go
cw9: hmm, why is instanceID confusing? also what is in DeploymentID?

tools/deploy/planner.go
```diff
@@ -0,0 +1,274 @@
...
+		for i, instance := range group.ToDeploy {
+			if !matchTargetType(instance.PlacementID, group.LeaderID, targetType) {
+				continue
+			}
+			target := deploymentTarget{
+				Instance:  instance,
+				Validator: p.validatorFactory.ValidatorFor(instance, group, targetType),
+			}
+			step.Targets = append(step.Targets, target)
+			group.removeInstanceToDeploy(i)
```
cw9: Is this function safe here? As you remove element from the slice, you swapped the last element to this index, and as you iterate to the next element, not sure if you would miss the last element since it's been moved around.

tools/deploy/helper.go
xichen2020: Hot spares are nice to haves but not a must-have given RF of 2. We can always provision some nodes as hot spares but don't deploy them until they are actually needed I assume?

tools/deploy/helper.go
xichen2020: So placement id is the instance id in the placement (e.g., `machine1:port1`), whereas deployment id is the instance id in the deployment (e.g., `machine1-service-appid`). The reason why we have both is that I don't want to enforce the constraint for all users of the tool that the instance id must be the same in the placement as the one in their deployment systems. Instead, the user can have different instance ids in the placement and their deployment system and provide a id conversion function to convert between them.

tools/deploy/planner.go
xichen2020: Hm yeah good catch, i'll write a unit test.

tools/deploy/planner.go
xichen2020: Sure I can record all errors. Probably can just use an error channel of size `len(group)` and process the errors at the end instead of introducing a locked multi error thing.

tools/deploy/planner.go
xichen2020: I see, the error message is different though, so all you save is the `select...` statement, I'd say probably not worth the effort.

tools/deploy/helper.go
xichen2020: Renamed them to `placementInstanceID` and `deploymentInstanceID` FYI.

tools/deploy/planner.go
prateek: Cool

tools/deploy/planner.go
prateek: Cool, SG

tools/deploy/helper.go
prateek: Cool, SG

tools/deploy/helper.go
xichen2020: Valid point, I added a `GenerateOneStep` function to the Planner API. I still generate the plan upfront for now because in practice we'll have no more than a few steps during a deploy so incrementally generating steps doesn't matter as much but we can easily change it if needed in the future with the `GenerateOneStep` API.

tools/deploy/helper.go
prateek: hey still planning on moving this out of the if?

tools/deploy/helper.go
```diff
@@ -0,0 +1,427 @@
...
+	return nil
+}
+
+func (h helper) executeStep(
+	step deploymentStep,
+	revision string,
+	all instanceMetadatas,
+) error {
+	h.logger.Infof("waiting until safe to deploy for step %v", step)
+	if err := h.waitUntilSafe(all); err != nil {
```
prateek: This is ensuring the entire cluster is healthy before deploying to a subset of nodes. Any reason to not restrict the check to only the nodes being deployed too?

xichen2020: Yeah so the rationale is to essentially wait for all the agg tier nodes to be healthy and not actively deployed to before initiating another deploy. This is primarily for extra caution in case  there are some nodes that have gone down after they are newly deployed, in which case we probably want to investigate before proceeding with a new round of deploy.

tools/deploy/helper.go
```diff
@@ -0,0 +1,448 @@
...
+			continue
+		}
+		filtered = append(filtered, metadata)
+	}
+	return filtered
+}
+
+// instanceMetadata contains instance metadata.
+type instanceMetadata struct {
+	// PlacementInstanceID is the instance id in the placement.
```
prateek: nice, these comments help a lot.

tools/deploy/helper.go
prateek: fair enough, paranoia ftw 👍

tools/deploy/helper.go
```diff
@@ -0,0 +1,434 @@
...
+	ForceMode
+)
+
+// Helper is a helper class handling deployments.
+type Helper interface {
+	// Deploy deploys a target revision to the instances in the placement.
+	Deploy(revision string, mode Mode) error
+}
+
+// TODO(xichen): disable deployment while another is ongoing.
```
xichen2020: Not fully, if you run the tool while another one is ongoing and is in between steps, you might still be able to proceed.

PR #41, opened 2017-08-22
Aggregator logic to support graceful topology changes

xichen2020:

This PR adds logic to the aggregation tier to support shard-level cutover and cutoff to enable graceful topology changes.


aggregator/aggregator.go
```diff
@@ -223,9 +220,6 @@ func (agg *aggregator) Open() error {
 	if agg.state != aggregatorNotOpen {
 		return errAggregatorAlreadyOpenOrClosed
 	}
-	if err := agg.placementWatcher.Watch(); err != nil {
```
jeromefroe: We don't need to watch the placement anymore?

aggregator/aggregator.go
```diff
@@ -369,48 +363,45 @@ func (agg *aggregator) updateShardsWithLock(newPlacement services.Placement) err
...
-		newShardLen := newShardIDs[len(newShardIDs)-1] + 1
+
+	// NB(xichen): shards are guaranteed to be sorted by their ids in ascending order.
+	var (
+		newShards   = newShardSet.All()
+		newShardIDs []uint32
+	)
+	if numShards := len(newShards); numShards > 0 {
+		newShardIDs = make([]uint32, 0, numShards)
+		newShardLen := newShards[numShards-1].ID() + 1
```
jeromefroe: Is the `+1` here just so we can do `incoming[shardID]` instead of `incoming[shardID-1]`? Do the shard ID's begin at 1 then instead of 0?

aggregator/aggregator.go
```diff
@@ -436,14 +427,40 @@ func (agg *aggregator) checkMetricType(mu unaggregated.MetricUnion) error {
...
 }
 
-func (agg *aggregator) ownedShards() []*aggregatorShard {
-	agg.RLock()
-	ownedShards := make([]*aggregatorShard, len(agg.shardIDs))
-	for i, shardID := range agg.shardIDs {
-		ownedShards[i] = agg.shards[shardID]
+func (agg *aggregator) ownedShards() (owned, toClose []*aggregatorShard) {
+	agg.Lock()
+	defer agg.Unlock()
```
jeromefroe: nit: seems this function can only returns at the end of the function body, perhaps close the lock explicitly before the return statement to save the overhead of `defer` if this in the hot path?

aggregator/aggregator_test.go
```diff
@@ -229,42 +247,109 @@ func TestAggregatorTick(t *testing.T) {
...
+			SetState(shard.Initializing).
+			SetCutoverNanos(0).
+			SetCutoffNanos(math.MaxInt64)
 	}
-	return testStagedPlacementProtoWithCustomShards(t, shardSet, testPlacementCutover)
+	return testStagedPlacementProtoWithCustomShards(t, instanceID, shardSet, testPlacementCutover)
 }
 
 func testStagedPlacementProtoWithCustomShards(
 	t *testing.T,
```
jeromefroe: nit: seems `t` isn't used in this function, perhaps remove it?

aggregator/flush_mgr.go
```diff
@@ -210,7 +205,22 @@ func (mgr *flushManager) Close() error {
 	return nil
 }
 
-func (mgr *flushManager) bucketForWithLock(l PeriodicFlusher) (*flushBucket, error) {
+func (mgr *flushManager) findOrCreateBucketWithLock(l PeriodicFlusher) (*flushBucket, error) {
+	bucket, err := mgr.findBucketWithLock(l)
+	if err == nil || err != errBucketNotFound {
```
jeromefroe: nit: seems odd to me to combine these two cases (though I can see why its valid). Perhaps it would cleaner to separate them out?

```
if err == nil {
  return bucket, nil
} else if err != errBucketNotFound {
  return nil, err
}
```

aggregator/flush_mgr.go
```diff
@@ -304,3 +308,15 @@ func newBucket(interval time.Duration, scope tally.Scope) *flushBucket {
 func (b *flushBucket) Add(flusher PeriodicFlusher) {
 	b.flushers = append(b.flushers, flusher)
 }
+
+func (b *flushBucket) Remove(flusher PeriodicFlusher) error {
```
jeromefroe: nit: seems `flushBucket` is not goroutine safe, perhaps it's worth adding that to it's comment on line 294?

aggregator/aggregator.go
xichen2020: I moved this call to `config/aggregator.go` since the watcher is shared between the aggregator and the flush manager.

aggregator/aggregator.go
xichen2020: This is because the shard ids start at 0, so the length needs to be 1 larger than the max id.

aggregator/aggregator.go
xichen2020: I did that initially but feels more explicit this way, thankfully this is only called once a tick which happens by default every 10 minutes.

aggregator/flush_mgr.go
xichen2020: Sure sounds good.

aggregator/flush_mgr.go
xichen2020: Sure sounds good, the bucket is always protected by the flush manager lock, can add a comment.

aggregator/list.go
```diff
@@ -182,34 +180,37 @@ func (l *metricList) PushBack(value interface{}) (*list.Element, error) {
...
-	// NB(xichen): it is important to determine ticking start time within the time lock
-	// because this ensures all the actions before `start` have completed if those actions
-	// are protected by the same read lock.
-	l.timeLock.Lock()
-	consumeStart := l.nowFn()
-	resolution := l.resolution
-	l.timeLock.Unlock()
-	alignedStartNanos := consumeStart.Truncate(resolution).UnixNano()
-	l.flushBefore(alignedStartNanos, consumeType)
+	// Unregisters the list outside the list lock to avoid holding the list lock
```
jeromefroe: nit: s/Unregisters/Unregister

aggregator/list_test.go
```diff
@@ -57,23 +58,89 @@ func TestMetricListPushBack(t *testing.T) {
...
+		flushTypeRes   flushType
+	)
+	opts := testOptions().SetClockOptions(clock.NewOptions().SetNowFn(nowFn))
+	l := newMetricList(testShard, time.Second, opts)
+	l.flushBeforeFn = func(beforeNanos int64, flushType flushType) {
+		beforeNanosRes = beforeNanos
+		flushTypeRes = flushType
+	}
+
+	for _, input := range []struct {
```
jeromefroe: nit: perhaps move the declaration of `input` the before the for loop?

aggregator/shard.go
```diff
@@ -87,3 +147,11 @@ func (s *aggregatorShard) Close() {
 	s.closed = true
 	s.metricMap.Close()
 }
+
+func (s *aggregatorShard) checkWritable() error {
```
jeromefroe: nit: perhaps `checkWritableWithLock` would be more appropriate since access to `s.latestWriteableNanos` is guarded with a lock?

aggregator/shard.go
```diff
@@ -49,16 +72,48 @@ func newAggregatorShard(shard uint32, opts Options) *aggregatorShard {
...
+	}
+
+	s.Lock()
+	s.earliestWritableNanos = earliestNanos
+	s.latestWriteableNanos = latestNanos
+	s.Unlock()
+}
+
+func (s *aggregatorShard) IsCutoff() bool {
+	nowNanos := s.nowFn().UnixNano()
```
jeromefroe: Do we want to get `nowNanos` while holding the lock to ensure calls to `IsCutoff` are synchronized with respect to calls to `AddMetricWithPoliciesList`?

aggregator/shard_test.go
```diff
@@ -21,19 +21,107 @@
...
+		require.Equal(t, input.expectedLatestNanos, shard.latestWriteableNanos)
+	}
+}
+
+func TestAggregatorShardIsCutoff(t *testing.T) {
+	now := time.Unix(0, 12345)
+	shard := newAggregatorShard(testShard, testOptions())
+	shard.nowFn = func() time.Time { return now }
+
+	for _, input := range []struct {
```
jeromefroe: nit: perhaps move the declaration of `inputs` above the for loop here as well (like in `TestAggregatorShardSetWritableRange`)?

aggregator/shard_test.go
```diff
@@ -21,19 +21,107 @@
...
+	err := shard.AddMetricWithPoliciesList(testValidMetric, testPoliciesList)
+	require.Equal(t, errAggregatorShardClosed, err)
+}
+
+func TestAggregatorShardAddMetricWithPoliciesListShardNotWriteable(t *testing.T) {
+	now := time.Unix(0, 12345)
+	shard := newAggregatorShard(testShard, testOptions())
+	shard.nowFn = func() time.Time { return now }
+
+	for _, input := range []struct {
```
jeromefroe: nit: same comment as above, maybe move this above the for loop?

aggregator/aggregator.go
jeromefroe: Gotcha, yea disregard this comment, was too early I guess :)

aggregator/aggregator.go
jeromefroe: Sounds good to me, I actually prefer `defer` myself so if it's called infrequently I like this approach.

aggregator/list.go
xichen2020: Sounds good.

aggregator/list_test.go
xichen2020: Hm why? it's only scoped to the for loop.

aggregator/shard.go
xichen2020: Don't think that's necessary, also this is called from the ticking goroutine so even if we hold a write lock here, once the call returns the goroutine no longer holds the lock so it's no different from the current implementation.

aggregator/shard.go
xichen2020: Right yeah sounds good.

aggregator/shard_test.go
xichen2020: I could, but I'd actually prefer to keep it scoped to the for loop.

aggregator/shard_test.go
xichen2020: Same as above.

aggregator/list_test.go
jeromefroe: Just seems cleaner to me, feels kind of cluttered with it in the for loop. Just a nit though, feel free to disregard.

xichen2020: Hm yeah, just looked through the go doc, seems moving it out of the loop is the recommended way (https://golang.org/doc/code.html#Testing), will do.

aggregator/aggregator.go
cw9: nit: Would be clearer to do 
`
maxShardID := newShards[numShards-1].ID()

incoming = make([]*aggregatorShard, maxShardID + 1)
`

aggregator/aggregator.go
```diff
@@ -124,7 +123,7 @@ type aggregatorMetrics struct {
 	gauges                    tally.Counter
 	invalidMetricTypes        tally.Counter
 	addMetricWithPoliciesList instrument.MethodMetrics
-	shards                    aggregatorShardMetrics
+	shards                    aggregatorShardsMetrics
```
cw9: rename shards to shardMetrics? or metricPerShard?

aggregator/flush_mgr.go
cw9: you can also get rid of that else

aggregator/leader_flush_mgr.go
```diff
@@ -282,20 +322,43 @@ func (t *leaderFlushTask) Run() {
...
+	var (
+		wgWorkers sync.WaitGroup
+		start     = mgr.nowFn()
+	)
+	for _, flusher := range t.flushers {
+		// By default traffic is cut off from a shard, unless the shard is in the list of
+		// shards owned by the instance, in which case the cutover time and the cutoff time
+		// are set to the corresponding cutover and cutoff times of the shard.
+		var cutoverNanos, cutoffNanos int64
+		if shards != nil {
```
cw9: could shards ever be nil?

aggregator/list.go
```diff
@@ -182,34 +180,37 @@ func (l *metricList) PushBack(value interface{}) (*list.Element, error) {
...
+		l.log.Errorf("error unregistering list: %v", err)
+	}
+}
 
-	flushDuration := l.nowFn().Sub(flushStart)
-	l.metrics.flushDuration.Record(flushDuration)
+func (l *metricList) Flush(cutoverNanos, cutoffNanos int64) {
+	start := l.nowFn()
+	alignedNowNanos := l.alignedNowNanos()
+	if alignedNowNanos > cutoverNanos && alignedNowNanos <= cutoffNanos {
```
cw9: so we are flushing (cutover, cutoff]? should it be [cutover, cutoff]?

aggregator/flush_mgr.go
```diff
@@ -304,3 +308,15 @@ func newBucket(interval time.Duration, scope tally.Scope) *flushBucket {
...
+	numFlushers := len(b.flushers)
+	for i := 0; i < numFlushers; i++ {
+		if b.flushers[i] == flusher {
+			b.flushers[i], b.flushers[numFlushers-1] = b.flushers[numFlushers-1], b.flushers[i]
+			b.flushers = b.flushers[:numFlushers-1]
```
martin-mao: Flushers don't async flush themselves yeah? Just checking we don't need to do anything with the flushers like close them or anything (I don't see anything in the interface)

aggregator/leader_flush_mgr.go
```diff
@@ -257,6 +274,29 @@ func (mgr *leaderFlushManager) persistFlushTimes(flushTimes *schema.ShardSetFlus
 	mgr.metrics.flushTimesPersist.ReportSuccessOrError(persistErr, mgr.nowFn().Sub(persistStart))
 }
 
+func (mgr *leaderFlushManager) ownedShards() (shard.Shards, error) {
+	stagedPlacement, onStagedPlacementDoneFn, err := mgr.placementWatcher.ActiveStagedPlacement()
+	if err != nil {
+		mgr.metrics.activeStagedPlacementErrors.Inc(1)
+		return nil, err
+	}
+	defer onStagedPlacementDoneFn()
```
martin-mao: Do we still want to call the onStagedPlacementDone function if line 285 fails? Does it need to do anything different if further actions fail?

aggregator/leader_flush_mgr.go
```diff
@@ -282,20 +322,43 @@ func (t *leaderFlushTask) Run() {
...
+		// By default traffic is cut off from a shard, unless the shard is in the list of
+		// shards owned by the instance, in which case the cutover time and the cutoff time
+		// are set to the corresponding cutover and cutoff times of the shard.
+		var cutoverNanos, cutoffNanos int64
+		if shards != nil {
+			shardID := flusher.Shard()
+			if shard, found := shards.Shard(shardID); found {
+				cutoverNanos = shard.CutoverNanos()
+				cutoffNanos = shard.CutoffNanos()
+			}
```
martin-mao: Isn't something bad going to happen if the shard is not found? Can you still flush with cutOver and cutOff set to 0?

aggregator/list.go
```diff
@@ -61,22 +62,17 @@ func newMetricListMetrics(scope tally.Scope) metricListMetrics {
 		flushCollected:      flushScope.Counter("collected"),
 		flushSuccess:        flushScope.Counter("success"),
 		flushErrors:         flushScope.Counter("errors"),
-		flushDuration:       flushScope.Timer("duration"),
 		flushStale:          flushScope.Counter("stale"),
-		flushBeforeDuration: scope.Timer("flush-before.duration"),
 		flushDiscarded:      flushScope.Counter("discarded"),
+		flush:               scope.Counter("flush"),
```
martin-mao: How is this different to the total of success and errors? Also a bit weird to have a count with name flush and a scope with the same prefix. Why aren't these all under a flush scope?

aggregator/list.go
```diff
@@ -182,34 +180,37 @@ func (l *metricList) PushBack(value interface{}) (*list.Element, error) {
...
 
-	flushDuration := l.nowFn().Sub(flushStart)
-	l.metrics.flushDuration.Record(flushDuration)
+func (l *metricList) Flush(cutoverNanos, cutoffNanos int64) {
+	start := l.nowFn()
+	alignedNowNanos := l.alignedNowNanos()
+	if alignedNowNanos > cutoverNanos && alignedNowNanos <= cutoffNanos {
+		l.flushBeforeFn(alignedNowNanos, consumeType)
+	} else {
+		l.flushBeforeFn(alignedNowNanos, discardType)
```
martin-mao: No metric required for discarding?

aggregator/map.go
```diff
@@ -123,6 +123,10 @@ func (m *metricMap) Tick(target time.Duration) tickResult {
 	}
 }
 
+func (m *metricMap) Close() {
+	m.metricLists.Close()
```
martin-mao: No need to protect against multiple or concurrent close of this list?

aggregator/aggregator.go
```diff
@@ -369,48 +363,45 @@ func (agg *aggregator) updateShardsWithLock(newPlacement services.Placement) err
 	}
 
 	var (
-		newShards = instance.Shards()
-		incoming  []*aggregatorShard
-		closing   []*aggregatorShard
+		newShardSet = instance.Shards()
+		incoming    []*aggregatorShard
```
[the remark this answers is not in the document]
xichen2020: Sure I can preallocate based on some high-level estimates there.

aggregator/aggregator.go
```diff
@@ -369,48 +363,45 @@ func (agg *aggregator) updateShardsWithLock(newPlacement services.Placement) err
...
+	for _, shard := range newShards {
+		shardID := shard.ID()
+		newShardIDs = append(newShardIDs, shardID)
 		if int(shardID) < len(agg.shards) && agg.shards[shardID] != nil {
 			incoming[shardID] = agg.shards[shardID]
 		} else {
 			incoming[shardID] = newAggregatorShard(shardID, agg.opts)
 			agg.metrics.shards.add.Inc(1)
 		}
+		incoming[shardID].SetWriteableRange(shard.CutoverNanos(), shard.CutoffNanos())
```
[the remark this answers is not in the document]
xichen2020: Sure I can define a range struct here, the xtime.Range struct in `m3x` has both fields as times so that won't work here.

aggregator/aggregator.go
```diff
@@ -436,14 +427,40 @@ func (agg *aggregator) checkMetricType(mu unaggregated.MetricUnion) error {
...
+}
+
+// closeShardsAsync asynchronously closes the shards to avoid blocking writes.
+// Because each shard write happens while holding the shard read lock, the shard
+// may only close itself after all its pending writes are finished.
+func (agg *aggregator) closeShardsAsync(shards []*aggregatorShard) {
+	for _, shard := range shards {
+		shard := shard
+		go func() {
+			shard.Close()
```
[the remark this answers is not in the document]
xichen2020: Sure I can emit a gauge recording the number of currently shards pending closing.

aggregator/flush_mgr.go
xichen2020: Yeah flushers are passive and their flushing is controlled by the flush manager. There is no need to close them or clean them up. These flushers are essentially lists storing aggregated metrics so they shouldn't be closed after flushing anyway.

aggregator/leader_flush_mgr.go
xichen2020: Yes `onStagedPlacementDone` needs to be called as long as the `.ActiveStagedPlacement()` call succeeds, regardless of whether further actions fail or not.

aggregator/leader_flush_mgr.go
xichen2020: Not necessarily. This may happen if the current instance no longer owns this shard (i.e., the shard is removed from the instance's shard list in the placement) and we are in the middle of asynchronously closing the removed shards (which removes the flusher from the list of flushers) but aren't done yet. Once the shards are closed, it should no longer hit this part again.

aggregator/list.go
xichen2020: Can get rid of the flush counter and use `flushSuccess` and `flushErrors` instead. Sure can consolidate the metric scope too.

aggregator/list.go
xichen2020: Can add metric sure.

aggregator/map.go
xichen2020: This is already handled in `metricsLists.Close`. It's safe to call it twice with the second call being a no op.

aggregator/aggregator.go
xichen2020: Can rename to `shardsMetrics` sure thing.

aggregator/aggregator.go
xichen2020: Sure can do.

aggregator/flush_mgr.go
xichen2020: Sure

aggregator/leader_flush_mgr.go
xichen2020: Yeah could probably simplify it be handling errors earlier, will do that.

aggregator/list.go
xichen2020: Yeah so the rationale here is that if `alignedNowNanos == cutoverNanos`, we should discard everything before cutoverNanos because the metrics before `cutoverNanos` have not been cut over yet so should be discarded, and if `alignedNowNanos == cutoffNanos`, we should still flush them because metrics before `cutoffNanos` have not been cut off yet so should still be flushed.

cw9: I see, would be good if you could put add some comments for this.

xichen2020: Sure thing.

aggregator/aggregator.go
xichen2020: Actually it's already inside an `aggregatorShardsMetrics` struct so when referenced it 'll look like `agg.metrics.shards.add.Inc(1)`, which seemed fairly self-explanatory and `agg.metrics.shardsMetrics.add.Inc(1)` seems a bit too verbose. I'll keep it as is for now and let me know if you think otherwise.

martin-mao:
LGTM

aggregator/aggregator.go
cw9: Sounds good

aggregator/aggregator.go
```diff
@@ -361,48 +359,49 @@ func (agg *aggregator) updateShardsWithLock(newPlacement services.Placement) err
...
 	}
-	// NB(xichen): shard ids are guaranteed to be sorted in ascending order.
-	newShardIDs := newShards.AllIDs()
-	if len(newShardIDs) > 0 {
-		newShardLen := newShardIDs[len(newShardIDs)-1] + 1
-		incoming = make([]*aggregatorShard, newShardLen)
-	}
-	for _, shardID := range newShardIDs {
+
+	// NB(xichen): shards are guaranteed to be sorted by their ids in ascending order.
```
jeromefroe: What guarantees that the shards are sorted here? Does it make sense to be defensive here and sort the shards just in case? If the sorting is truly unnecessary perhaps add a comment explaining why?

xichen2020: It's guaranteed by the `Shards()` implementation. I could create an issue in m3cluster to add comments saying the returned shards are always sorted by id in ascending order?

aggregator/aggregator.go
```diff
@@ -428,14 +427,45 @@ func (agg *aggregator) checkMetricType(mu unaggregated.MetricUnion) error {
...
+	agg.Lock()
+	defer agg.Unlock()
+
+	owned = make([]*aggregatorShard, 0, len(agg.shardIDs))
+	for i := 0; i < len(agg.shardIDs); i++ {
+		shardID := agg.shardIDs[i]
+		// NB(xichen): if traffic has been cut off for a shard, the aggregator actively
+		// closes it to avoid wasting CPU cycles flushing its metric lists.
+		if shard := agg.shards[shardID]; !shard.IsCutoff() {
+			owned = append(owned, shard)
```
jeromefroe: nit: perhaps add a `continue` at the end of the if clause so that we don't need the else clause?

aggregator/aggregator.go
```diff
@@ -428,14 +427,45 @@ func (agg *aggregator) checkMetricType(mu unaggregated.MetricUnion) error {
...
+func (agg *aggregator) closeShardsAsync(shards []*aggregatorShard) {
+	atomic.AddInt32(&agg.shardsPendingClose, int32(len(shards)))
+	agg.metrics.shards.pendingClose.Update(float64(atomic.LoadInt32(&agg.shardsPendingClose)))
+
+	for _, shard := range shards {
+		shard := shard
+		go func() {
+			shard.Close()
+			atomic.AddInt32(&agg.shardsPendingClose, -1)
+			agg.metrics.shards.pendingClose.Update(float64(atomic.LoadInt32(&agg.shardsPendingClose)))
```
jeromefroe: nit: `atomic.AddInt32` returns the new value of the integer so we don't need the additional load here.

aggregator/flush_mgr.go
```diff
@@ -303,3 +315,15 @@ func newBucket(interval time.Duration, scope tally.Scope) *flushBucket {
 func (b *flushBucket) Add(flusher PeriodicFlusher) {
 	b.flushers = append(b.flushers, flusher)
 }
+
+func (b *flushBucket) Remove(flusher PeriodicFlusher) error {
+	numFlushers := len(b.flushers)
+	for i := 0; i < numFlushers; i++ {
+		if b.flushers[i] == flusher {
+			b.flushers[i], b.flushers[numFlushers-1] = b.flushers[numFlushers-1], b.flushers[i]
```
jeromefroe: nit: Any reason we're swapping both positions? Seems it would be sufficient to just move the last entry into index `i` and then shorten the length of the slice.

aggregator/list.go
```diff
@@ -182,39 +185,48 @@ func (l *metricList) PushBack(value interface{}) (*list.Element, error) {
...
-	l.metrics.flushDuration.Record(flushDuration)
+func (l *metricList) Flush(cutoverNanos, cutoffNanos int64) {
+	start := l.nowFn()
+	alignedNowNanos := l.alignedNowNanos()
+
+	// NB(xichen): If alignedNowNanos <= cutoverNanos, the aggregated metrics should
+	// be discarded because they are aggregated before traffic was cut over. On the
+	// other hand, if alignedNowNanos > cutoffNanos, the aggregated metrics should
+	// also be discarded because they are aggregated after traffic was cut off.
+	if alignedNowNanos > cutoverNanos && alignedNowNanos <= cutoffNanos {
```
jeromefroe: super nit: perhaps switch the case statements to match the comment? Currently the comment discusses the conditions for dropping metrics but the if statements checks if they can be consumed.

aggregator/options.go
```diff
@@ -95,52 +94,63 @@ var (
...
-	electionManager                ElectionManager
-	resignTimeout                  time.Duration
-	entryPool                      EntryPool
-	counterElemPool                CounterElemPool
-	timerElemPool                  TimerElemPool
-	gaugeElemPool                  GaugeElemPool
-	bufferedEncoderPool            msgpack.BufferedEncoderPool
-	aggTypesPool                   policy.AggregationTypesPool
-	quantilesPool                  pool.FloatsPool
+	defaultCounterAggregationTypes   policy.AggregationTypes
```
jeromefroe: super nit: this struct is pretty large, does it make sense to group some of the fields into their own smaller options and include them, kind of like we do with `clockOpts`, `intstrumentOpts`, and others? For example, it looks like the prefixes and suffixes could be grouped together in their own options type.

aggregator/aggregator.go
xichen2020: Actually I feel it's more clear this way, would prefer to keep it as is unless you feel strongly.

aggregator/aggregator.go
xichen2020: Sounds good.

aggregator/flush_mgr.go
xichen2020: Sure sounds good.

aggregator/list.go
xichen2020: Yeah I prefer to handle the common case first, will leave it as is unless you feel strongly.

aggregator/options.go
xichen2020: Possibly, although there are derived fields for performance reasons that are computed from the base fields and are updated when the base fields are set. Those derived fields will need to be grouped together w/ the base fields for that to continue to work. I can look at it in a separate refactoring PR.

aggregator/aggregator.go
jeromefroe: Probably not necessary, maybe just augment the comment to note that `shards returned from instance.Shards() are guaranteed to be sorted ...`

aggregator/aggregator.go
jeromefroe: Don't feel strongly at all, I think `continue` is arguably more idiomatic is the only reason I mentioned it.

aggregator/list.go
jeromefroe: Fine by me. Definitely a super nit :)

PR #45, opened 2017-08-27
Sort targets by instance id for deterministic ordering

xichen2020:
cc @cw9 @prateek @jeromefroe 

This PR sorts the deployment targets in each deployment step by their instance IDs for deterministic ordering so it's easier to see in logs.

### DIFF
--- a/tools/deploy/planner.go
+++ b/tools/deploy/planner.go
@@ -22,6 +22,7 @@ package deploy
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 
 	"github.com/m3db/m3cluster/services"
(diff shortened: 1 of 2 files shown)


tools/deploy/planner.go
```diff
@@ -135,6 +136,8 @@ func (p deploymentPlanner) generateStepFromTargetType(
 			delete(instances, shardSetID)
 			continue
 		}
+
+		done := false
```
jeromefroe: nit: use default initialization, ie `var done bool`

xichen2020: Yeah I tend to be more explicit in cases like this, feels more clear to the readers.

cw9: nit: it feels cleaner to create a generateSortedStepFromTargetType that sorts the step generated from generateStepFromTargetType

xichen2020: Hm that feels like a bit overkill, this is only sorting instances within the same step so whether they sorted or not doesn't affect correctness of the output, so might as well always sort them for more deterministic output.

cw9: sounds good

cw9:
LGTM

PR #50, opened 2017-09-11
Agg tier topoology change support

xichen2020:

This PR adds logic to gracefully handle agg tier topology changes and backing out of those changes. The notable changes include:
* Allow instances not in the placement to start up so we can run hot spares.
* Automatically start or stop campaigning based on placement information.

I'll work on adding more test coverage while the PR is being reviewed.

### DIFF
--- a/aggregator/election_mgr.go
+++ b/aggregator/election_mgr.go
@@ -382,6 +382,7 @@ func (mgr *electionManager) Resign(ctx context.Context) error {
 			return true
 		}
 	}
+	// Log the context error because the error returned from the retrier is not helpful.
 	if err := mgr.resignWhile(ctxNotDone); err != nil {
 		mgr.metrics.resignTimeout.Inc(1)
 		mgr.logError("resign error", ctx.Err())
@@ -413,6 +414,9 @@ func (mgr *electionManager) Close() error {
 	mgr.Unlock()
 
 	mgr.Wait()
+	mgr.campaignStateWatchable.Close()
+	mgr.electionStateWatchable.Close()
+	mgr.goalStateWatchable.Close()
 	return nil
 }
 
--- a/aggregator/election_mgr.go
+++ b/aggregator/election_mgr.go
@@ -26,11 +26,13 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/m3db/m3cluster/services"
 	"github.com/m3db/m3cluster/services/leader/campaign"
 	"github.com/m3db/m3x/clock"
+	"github.com/m3db/m3x/errors"
 	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/retry"
 	"github.com/m3db/m3x/watch"
@@ -40,11 +42,18 @@ import (
 
 // ElectionManager manages leadership elections.
 type ElectionManager interface {
+	// Reset resets the election manager.
+	Reset() error
+
 	// Open opens the election manager for a given shard set.
 	Open(shardSetID uint32) error
 
 	// ElectionState returns the election state.
 	ElectionState() ElectionState
+
+	// IsCampaigning returns true if the election manager is actively campaigning,
+	// and false otherwise.
+	IsCampaigning() bool
 
 	// Resign stops the election and resigns from the ongoing campaign if any, thereby
 	// forcing the current instance to become a follower. If the provided context
@@ -63,8 +72,9 @@ const (
 var (
 	errElectionManagerAlreadyOpenOrClosed = errors.New("election manager is already open or closed")
 	errElectionManagerNotOpenOrClosed     = errors.New("election manager is not open or closed")
-	errElectionManagerAlreadyResigning    = errors.New("election manager is already resigning")
+	errElectionManagerOpen                = errors.New("election manager is open")
 	errLeaderNotChanged                   = errors.New("leader has not changed")
+	errUnexpectedShardCutoverCutoffTimes  = errors.New("unexpected shard cutover and/or cutoff times")
 )
 
 type electionManagerState int
@@ -86,11 +96,11 @@ const (
 	// Follower state.
 	FollowerState
 
-	// Leader state.
-	LeaderState
-
 	// Pending follower state.
 	PendingFollowerState
+
+	// Leader state.
+	LeaderState
 )
 
 func newElectionState(state campaign.State) (ElectionState, error) {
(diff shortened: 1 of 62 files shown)


aggregator/election_mgr.go
```diff
@@ -420,22 +507,174 @@ func (mgr *electionManager) verifyPendingFollower(watch xwatch.Watch) {
...
+
+		if err := multiErr.FinalError(); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
+	// TODO(xichen): handle this to enable multiple concurrent topology changes.
+	mgr.metrics.campaignCheckUnexpectedShardTimes.Inc(1)
+	return false, errUnexpectedShardCutoverCutoffTimes
```
cw9: this is someCutover+someCutoff? Which is a possible valid state?

aggregator/flush_mgr.go
```diff
@@ -243,7 +280,7 @@ func (mgr *flushManager) findBucketWithLock(l PeriodicFlusher) (*flushBucket, er
 // when I have more time I'll spend a few hours to get timer.Reset() right and switch
 // to a timer.Start + timer.Stop + timer.Reset + timer.After based approach.
 func (mgr *flushManager) flush() {
-	defer mgr.wgFlush.Done()
+	defer mgr.Done()
```
cw9: reviewed until here, will continue tomorrow

aggregator/election_mgr.go
```diff
@@ -320,18 +400,23 @@ func (mgr *electionManager) Resign(ctx context.Context) error {
...
-	defer mgr.Unlock()
-
 	if mgr.state != electionManagerOpen {
+		mgr.Unlock()
 		return errElectionManagerNotOpenOrClosed
 	}
 	close(mgr.doneCh)
 	mgr.state = electionManagerClosed
+	mgr.Unlock()
+
```
cw9: close the watchables so whoever is listening on them can receive the final signals?

aggregator/follower_flush_mgr.go
```diff
@@ -114,51 +110,46 @@ func newFollowerFlushManager(
...
 		mgr.metrics.kvUpdateFlush.Inc(1)
 	} else {
 		durationSinceLastFlush := now.Sub(mgr.lastFlushed)
-		if durationSinceLastFlush > mgr.maxNoFlushDuration {
-			mgr.lastFlushed = now
-			shouldFlush = true
-			mgr.metrics.forcedFlush.Inc(1)
-			flushBeforeNanos := now.Add(-mgr.maxNoFlushDuration).Add(mgr.forcedFlushWindowSize).UnixNano()
+		if durationSinceLastFlush > mgr.maxBufferSize {
+			flushBeforeNanos := now.Add(-mgr.maxBufferSize).Add(mgr.forcedFlushWindowSize).UnixNano()
```
cw9: should only discard as late as now.Add(-mgr.maxBufferSize)

aggregator/placement_mgr.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return false, err
+	}
+	currInstance, err := mgr.instanceFrom(placement)
+	if err != nil {
+		return false, err
+	}
+	currShardSetID := currInstance.ShardSetID()
+	currShards := currInstance.Shards().All()
+	for _, currShard := range currShards {
```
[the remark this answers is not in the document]
xichen2020: Sounds good.

aggregator/placement_mgr.go
```diff
@@ -0,0 +1,239 @@
...
+			return false, nil
+		}
+	}
+	allInstances := placement.Instances()
+	for _, instance := range allInstances {
+		if instance.ShardSetID() != currShardSetID || instance.ID() == mgr.instanceID {
+			continue
+		}
+		otherShards := instance.Shards().All()
+		if len(otherShards) != len(currShards) {
```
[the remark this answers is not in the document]
xichen2020: I actually need all the shards down below so just saving a few extra calls here.

aggregator/election_mgr.go
```diff
@@ -244,64 +319,69 @@ func (mgr *electionManager) Open(shardSetID uint32) error {
...
+		select {
+		case <-ctx.Done():
+			return false
+		default:
+			return true
+		}
+	}
+	if err := mgr.resignWhile(ctxNotDone); err != nil {
+		mgr.metrics.resignTimeout.Inc(1)
+		mgr.logError("resign error", ctx.Err())
```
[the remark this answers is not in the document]
xichen2020: The error returned is `errWhileConditionFalse`, which is not helpful. Also if an error is returned, we know the context is done. Added a comment.

aggregator/placement_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3cluster/shard"
+	"github.com/m3db/m3x/clock"
+	"github.com/uber-go/tally"
+)
+
+var (
+	errPlacementManagerNotOpenOrClosed = errors.New("placement manager not open or closed")
+	errPlacementManagerOpenOrClosed    = errors.New("placement manager already open or closed")
+	errInstanceNotFoundInPlacement     = errors.New("instance not found in placement")
```
cw9: Should this error be public since it's being checked by the user of PlacementManager

aggregator/election_mgr_options.go
```diff
@@ -80,28 +90,60 @@ type ElectionManagerOptions interface {
...
+	SetCampaignStateCheckInterval(value time.Duration) ElectionManagerOptions
+
+	// CampaignStateCheckInterval returns the interval to check campaign state.
+	CampaignStateCheckInterval() time.Duration
+
+	// SetShardCutoffCheckOffset sets the offset when checking if a shard has been cutoff.
+	SetShardCutoffCheckOffset(value time.Duration) ElectionManagerOptions
+
+	// ShardCutoffCheckOffset returns the offset when checking if a shard has been cutoff.
+	ShardCutoffCheckOffset() time.Duration
```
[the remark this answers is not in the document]
cw9: I see, could you modify the comment to indicate that this offset is being used to stop campaigning before its real cutoff time? Cause right now reader needs to go to where this value is used to understand what this field is.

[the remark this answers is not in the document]
xichen2020: Sure thing.

aggregator/placement_mgr.go
xichen2020: Sure thing.

aggregator/follower_flush_mgr.go
```diff
@@ -150,11 +152,12 @@ func (mgr *followerFlushManager) Prepare(buckets []*flushBucket) (flushTask, tim
 		if mgr.flushMode == forcedFollowerFlush && durationSinceLastFlush >= mgr.forcedFlushWindowSize {
 			flushBeforeNanos := now.Add(-mgr.maxBufferSize).UnixNano()
 			flushersByInterval = mgr.flushersFromForcedFlush(buckets, flushBeforeNanos)
+			needsFlush = true
 			mgr.metrics.forcedFlush.Inc(1)
 		}
 	}
 
-	if len(flushersByInterval) == 0 {
+	if !needsFlush {
```
cw9: is this for readability or is there a case where len(flushersByInterval) == 0 does not equal to !needsFflush?

aggregator/election_mgr.go
```diff
@@ -496,8 +499,12 @@ func (mgr *electionManager) verifyPendingFollower(watch xwatch.Watch) {
 			}
 			return nil
 		}); verifyErr != nil {
-			mgr.logError("verify error", verifyErr)
-			continue
+			// If state has changed, we skip this stale change.
+			if !stateUnchangedFn() {
```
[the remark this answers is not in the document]
xichen2020: That should never happen, because if it's evaluated as false, it means there is a new state set on the watchable, therefore incrementing the state id, so it should always be false afterwards.

aggregator/follower_flush_mgr.go
xichen2020: So I changed it so that if the kv is updated, or it's time to do a forced flush, we'll flush and **update the flush time** even if len(flushersByInterval) == 0 (e.g., the instance doesn't own any shards). Otherwise if we don't update the time, the forced flush will be performed every second since `durationSinceLastFlushed` will keep increasing.

cw9: Got it, maybe add a comment for this?

aggregator/election_mgr.go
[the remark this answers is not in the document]
cw9: sounds good

PR #52, opened 2017-09-19
Agg tier supports partitioned backend

### LINKED ISSUE #31: Shard ids before flush
M3aggregator needs to be able to shard ids in order to support sharded backends.

xichen2020:

This PR adds logic to the agg tier to support partitioned backend (e.g., indexers and analyzers). Fixed #31. 

In the current implementation, each list encodes aggregated metric into an internal buffer during a flush regardless of which partition the metric belongs to, and once the buffer size exceeds a given threshold, it calls handler.Handle(buffer). The handler then is responsible for sending the buffer to backend services (e.g., ingesters). However, this won't work for partitioned backend in that each buffer passed to the handler contains metrics in different partitions.

**Update:** I decided to adopt a third alternative that provides cleaner API and `Handler` interface and better encapsulation without performance degradation even as we scale up the number of backends.

Notable changes include:
* The encoding logic has been moved to be purely contained in the writer implementation. The aggregator list simply calls `writer.Write(...)` passing in the aggregated metric alongside its policy without having to understand how the data is encoded, which provides better decoupling between flushing logic and the logic that encodes data and writes them to the backends.
* Reworked the `Handler` interface and introduced `Writer`, `Router`, and `Queue` and make them reusable. Client uses `Handler` to create new writers, which encodes data internally and flushes them when the internal buffer exceeds a threshold. The flushed buffer is then sent to the router, which takes care of routing the buffer to different backend queues, which eventually forwards the buffer to the backends.
* In order to make sure we don't encode data more times than necessary, we group the backends by their sharding functions so that if two backends share the same sharding function (including the hash type and the total number of shards), they belong to the same group which receives ref-counted buffers from a single writer so that aggregated metrics are only encoded once per sharding function to save CPU cycles.
* In order to reduce lock contention among different lists, the writers returned from the handlers are intentionally thread-unsafe. Each list holds on to a writer it creates when the list is created, and uses the writer to write data in a flush. Because a list only performs at most one flush at any given time, this is functionally correct, and because the lists writes data independently, they don't contend with each other, therefore providing a better story when we need to scale up the number of lists.

~~There are two approaches I considered:~~
* ~~Change the handler interface to `Handle(metric, policy)` and move the responsibility of encoding metrics alongside the policy from the lists to the handlers. This allows for a cleaner interface and better encapsulation, in that only the partitioned handler needs to know about how metrics are partitioned, and different partitioned handlers may use different partitioning functions. The downside of this approach is that because each handler encodes metrics independently, each metric + policy is therefore encoded N times, where N is the number of backends. As such, the more backends the agg tier has, the more times each metric is encoded, and because the encoding process is on the critical path of a flush, this also means the flush times degrade as the number of backends increases. Additionally, because the handler is centrally responsible for handling metrics from all lists during flushes, this could lead to increased lock contention when multiple lists flush at the same time, which also degrades flush times.~~

* ~~Alternatively, change the handler interface to `Handle(partition, buffer)`. The list is still responsible for encoding data, but in a partition-aware fashion. Once the buffer size for a given partition is big enough, the buffer is then passed along to the handler alongside the associated partition. Because the buffer is ref-counted, it's therefore reused by all backends, which means each metric + policy only needs to be encoded once regardless of the number of backends, providing a better scaling story. Additionally, because each list encodes data independently and only passes the buffer to the handler when the buffer is full, there's practically no point of contention among them. The downside of this approach is that the list now needs to be aware of partitions, and all partitioned handlers are required to use the same partitioning function (although the latter is a non-issue at the moment), leading to a slightly less flexible design.~~

~~This PR takes approach 2 to improve agg tier performance at the cost of slightly less flexible APIs, since the agg tier is currently CPU-bound and there is no immediate need for agg tier backends to customize partitioning functions on a per-backend basis. If such need arises in the medium to longer term, I can change the interface to allow more flexibility in backend configurations.~~

Feel free to take a look and leave comments.


aggregator/handler/config.go
```diff
@@ -137,10 +135,98 @@ func (c *forwardHandlerConfiguration) NewHandler(
...
+		serversAssigned    = make(map[string]struct{})
+		partitionsAssigned = make(map[int]struct{})
+	)
+	for _, partitions := range c.Partitions {
+		// Make sure we have a deterministic ordering.
+		sortedPartitions := make([]int, 0, len(partitions.Range))
+		for partition := range partitions.Range {
+			sortedPartitions = append(sortedPartitions, int(partition))
+		}
+		sort.Ints(sortedPartitions)
```
jeromefroe: Any reason not to implement `sort.Interface` for `BackendServerPartition` so we can sort `partitions` directly?

aggregator/handler/partitioned.go
```diff
@@ -0,0 +1,120 @@
...
+
+func newPartitionedHandlerMetrics(scope tally.Scope) partitionedHandlerMetrics {
+	return partitionedHandlerMetrics{
+		partitionNotOwned: scope.Counter("partition-not-owned"),
+	}
+}
+
+type handlerWithPartition struct {
+	Partitions partitioning.PartitionSet
+	Handler    aggregator.Handler
```
jeromefroe: nit: perhaps just embed `Handler` so users can call `handler.Handle` instead of `handler.Handler.Handle`?

aggregator/handler/partitioned.go
```diff
@@ -0,0 +1,120 @@
...
+	return &partitionedHandler{
+		handlers: handlers,
+		metrics:  newPartitionedHandlerMetrics(scope),
+	}, nil
+}
+
+func (h *partitionedHandler) Handle(buffer aggregator.PartitionedBuffer) error {
+	for _, handler := range h.handlers {
+		if handler.Partitions.Contains(buffer.Partition) {
+			return handler.Handler.Handle(buffer)
```
jeromefroe: Is the expectation that we will have a single `partitionedHandler` for each backend? For example, if we have three backends, ingesters, indexers, and analyzers, would we have a single `partitionedHandler` for each? Seems like this case should only occur if there was some kind of configuration error, i.e. a handler was constructed with 512 partitions instead of 1024 or something of that nature?

aggregator/list.go
```diff
@@ -94,22 +103,23 @@ type metricList struct {
...
-	closed             bool
-	encodeFn           encodeFn
-	flushBeforeFn      flushBeforeFn
-	consumeAggMetricFn aggMetricFn
-	discardAggMetricFn aggMetricFn
-	metrics            metricListMetrics
+	closed              bool
+	aggregations        *list.List
+	lastFlushedNanos    int64
+	encodersByPartition []msgpack.AggregatedEncoder
```
jeromefroe: Just a thought (admittedly could be totally off the mark here): would it make sense for the list to be aware of the partition ranges instead of the total number of partitions? In this way, `encodersByPartition` would be much smaller and we would be able to more easily support a much higher number of partitions. Also, since the handlers are per partition range, it would be consistent with what they expect.

partitioning/partitions.go
```diff
@@ -0,0 +1,155 @@
...
+
+var (
+	// Partition range is expected to provided in the form of startPartition..endPartition.
+	// An example partition range is 0..63.
+	rangeRegexp = regexp.MustCompile(`^([0-9]+)(\.\.([0-9]+))?$`)
+
+	errInvalidPartition = errors.New("invalid partition")
+)
+
+// A PartitionSet is a range of partitions organized as a set.
```
jeromefroe: nit: s/A PartitionSet/PartitionSet

partitioning/partitions.go
```diff
@@ -0,0 +1,155 @@
...
+	"strconv"
+)
+
+const (
+	defaultNumPartitions = 32
+)
+
+var (
+	// Partition range is expected to provided in the form of startPartition..endPartition.
+	// An example partition range is 0..63.
```
jeromefroe: Perhaps clarify if the min and max are inclusive or exclusive?

partitioning/partitions.go
```diff
@@ -0,0 +1,155 @@
...
+			case int:
+				ps.Add(uint32(c))
+			default:
+				return fmt.Errorf("unexpected range %v", c)
+			}
+		}
+		return nil
+	}
+
+	// Otherwise try to parse out a single partition.
```
jeromefroe: Any reason we don't require that everything be of the form `x..y`? A single partition could be represented at `x..x`, assuming the bounds are inclusive. If we do want to allow single integers though perhaps update the comment for `PartitionSet` to reflect that.

aggregator/handler/config.go
```diff
@@ -137,10 +135,98 @@ func (c *forwardHandlerConfiguration) NewHandler(
...
+		opts = opts.SetQueueSize(c.QueueSize)
+	}
+	return NewPartitionedHandler(c.Partitions, opts)
+}
+
+// validatePartitions ensures a single partition isn't present multiple times, and
+// that a single server isn't assigned to multiple partition ranges.
+func (c *partitionedHandlerConfiguration) validatePartitions() error {
+	var (
+		serversAssigned    = make(map[string]struct{})
```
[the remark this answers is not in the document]
xichen2020: Can do, might be a bit overkill though given that this is called once at startup time.

aggregator/handler/config.go
xichen2020: Not sure what you meant by "sort partitions directly", internally the `backendServerPartition` has a partition set which is a map[uint32]struct{}, this is to sort the partitions contained in the set to have deterministic ordering when validating them.

aggregator/handler/partitioned.go
xichen2020: Sure can probably embed both.

aggregator/handler/partitioned.go
xichen2020: Yes, there is one `partitionedHandler` for each partitioned backend (indexers and analyzers). As you said, this is essentially to catch cases where some partitions are not assigned to any servers so we'd probably want to know when that happens.

aggregator/list.go
xichen2020: That seems to cause tighter coupling between the partitioned handler and the list in that now the list needs to decide which partition range a metric belongs to. The interface would also look a bit weird (e.g., the `Handle` function would take a partition range instead of a partition).

I'm not too concerned about the overhead to keep encoders per partition, given 16 shards per instance, 1K partitions, 1.44KB max flush size, the space taken by the encoders is 16 * 1K * 1.44KB = 20MB, which is quite reasonable.

partitioning/partitions.go
xichen2020: Sure thing.

partitioning/partitions.go
xichen2020: Sounds good.

partitioning/partitions.go
xichen2020: To support non-contiguous ranges and single partition more naturally. Can update the comment sure thing.

aggregator/handler/config.go
jeromefroe: Ahh, disregard, I thought `partitions.Range` was a slice for some reason.

aggregator/list.go
jeromefroe: Sounds good, wasn't totally sold on the idea, just figured I would just play devil's advocate.

aggregator/handler/handler.go
```diff
@@ -20,13 +20,55 @@
...
+
+// UnmarshalYAML unmarshals YAML into a type.
+func (t *Type) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var str string
+	if err := unmarshal(&str); err != nil {
+		return err
+	}
+	if str == "" {
+		*t = unknownType
+		return errUnknownHandlerType
```
cw9: nit: seems this should be errNoHandlerType or we could just skip this check and use the err returned in the end of this function.

aggregator/list.go
```diff
@@ -37,6 +38,10 @@ import (
 	"github.com/uber-go/tally"
 )
 
+const (
+	defaultNumPartitions = 1024
```
cw9: nit: move this to Options?

aggregator/handler/partitioned.go
```diff
@@ -0,0 +1,120 @@
...
+	var (
+		// Divide overall queue size among all partitions.
+		partitionQueueSize = opts.QueueSize() / numPartitions
+		handlers           = make([]handlerWithPartition, 0, numPartitions)
+		instrumentOpts     = opts.InstrumentOptions()
+		scope              = instrumentOpts.MetricsScope()
+	)
+	for _, partition := range partitions {
+		handlerScope := scope.Tagged(map[string]string{
+			"min-partition": strconv.Itoa(partition.Range.Min()),
```
cw9: If we start to put non-consecutive partitions in a partitionset then these 2 fields in this log would be not very helpful, how about we add a name/label to each partition and use the label for metrics/logging? 

For example if we createa a partition with range (e.g "0..50") then we use the range as the label, if we use dynamically generated m3cluster/placement, then we can perhaps use shardSetID as the label.

I imagine one day we might start supporting dynamic backends for indexers to save some deployments on the agg tier, so we should take that into consideration and try to stay closer to generic m3cluster placements.

aggregator/types.go
```diff
@@ -25,16 +25,23 @@ import (
...
 	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3x/pool"
 )
 
-// ShardFn maps a id to a shard given the total number of shards.
+// ShardFn maps a id to an aggregator shard given the total number of shards.
 type ShardFn func(id []byte, numShards int) uint32
 
+// PartitionFn maps a chunked id to a backend partition.
+type PartitionFn func(chunkedID id.ChunkedID) uint32
```
cw9: I know we use partition for these kinds of use cases, but can we unify them to "shard" in m3db repos?

aggregator/handler/partitioned.go
```diff
@@ -0,0 +1,120 @@
...
+		})
+	}
+	return &partitionedHandler{
+		handlers: handlers,
+		metrics:  newPartitionedHandlerMetrics(scope),
+	}, nil
+}
+
+func (h *partitionedHandler) Handle(buffer aggregator.PartitionedBuffer) error {
+	for _, handler := range h.handlers {
```
cw9: can we have a flat list of handlers with length of total number of partitions we support? and directly ref to the forwardHandler for that partition? To save the iteration and map look up here?

services/m3aggregator/config/aggregator.go
```diff
@@ -160,6 +163,12 @@ type AggregatorConfiguration struct {
 	// Maximum flush size in bytes.
 	MaxFlushSize int `yaml:"maxFlushSize"`
 
+	// Total number of backend partitions.
+	TotalPartitions int `yaml:"totalPartitions" validate:"nonzero"`
+
+	// Partition function type.
+	PartitionFnType *hashFnType `yaml:"partitionFnType"`
```
cw9: rename to [Backend|Flush]ShardFnType?

tools/deploy/validator_test.go
```diff
@@ -193,12 +193,6 @@ func TestValidatorForLeaderFollowerSuccess(t *testing.T) {
 	require.NoError(t, validator())
 }
 
-type capturingValidator struct {
```
cw9: is this change related?

aggregator/handler/handler.go
xichen2020: Sure can skip the check.

aggregator/handler/partitioned.go
xichen2020: Sure can add a `Desc` or a `Name` to the `BackendServerPartition` struct and use that as the tag here.

aggregator/handler/partitioned.go
xichen2020: Sure can do, sounds good.

aggregator/list.go
xichen2020: This is mostly just an optimization to allocate some initial capacity, doubt we'd ever need to change it so probably not worth it to make it optional.

aggregator/types.go
xichen2020: Can take a look to unify the naming convention, I intentionally called it something else to distinguish agg tier shards and backend shards, but that's fair too.

services/m3aggregator/config/aggregator.go
xichen2020: I'll think of a more appropriate name.

tools/deploy/validator_test.go
xichen2020: Not directly related, one of the deployment planner tests was flaky so this was just to fix that flaky test.

services/m3aggregator/config/aggregator.go
```diff
@@ -531,50 +551,72 @@ func (c *kvClientConfiguration) NewKVClient(instrumentOpts instrument.Options) (
...
+
+func (t hashFnType) PartitionFnGen(totalPartitions int) (aggregator.PartitionFnGen, error) {
+	switch t {
+	case murmur32HashFn:
+		// NB(xichen): If this turns out to be too CPU intensive due to byte copies, can rewrite
+		// it to compute murmur3 hashes with zero byte copies.
+		return func() aggregator.PartitionFn {
+			// Create a new local buffer on every function invocation, therefore
+			// making the generated partition functions independent of each other.
+			buf := bytes.NewBuffer(make([]byte, 0, initialChunkedIDSize))
```
jeromefroe: literally just read a blog post that might be relevant here: https://segment.com/blog/allocation-efficiency-in-high-performance-go-services/

I haven't done any benchmarks but I think it might be more efficient to do the following:

```
return func(chunkedID id.ChunkedID) uint32 {
    var b [256]byte
    buf := b[:0]

    buf = append(buf, chunkedID.Prefix...)
    buf = append(buf, chunkedID.Data...)
    buf = append(buf, chunkedID.Suffix...)
    return murmur3.Sum32(buf) % uint32(totalPartitions)
}
```

This *should* only allocate when the total ID length is greater than 256. Need to run the benchmarks to confirm.

jeromefroe: Huzzah! It worked!

Benchmark:

```
package main

import (
	"bytes"
	"testing"

	"github.com/spaolacci/murmur3"
)

var (
	foo = []byte("foo")
	bar = []byte("bar")
	baz = []byte("baz")
)

func BenchmarkBuffer(b *testing.B) {
	var hash uint32
	for n := 0; n < b.N; n++ {
		buf := bytes.NewBuffer(make([]byte, 0, 16))
		buf.Write(foo)
		buf.Write(bar)
		buf.Write(baz)
		hash = murmur3.Sum32(buf.Bytes())
	}
	_ = hash
}

func BenchmarkArray(b *testing.B) {
	var hash uint32
	for n := 0; n < b.N; n++ {
		var a [16]byte
		buf := a[:0]
		buf = append(buf, foo...)
		buf = append(buf, bar...)
		buf = append(buf, baz...)
		hash = murmur3.Sum32(buf)
	}
	_ = hash
}
```

Result:

```
BenchmarkBuffer-4       10000000               123 ns/op             128 B/op          2 allocs/op
BenchmarkArray-4        50000000                24.9 ns/op             0 B/op          0 allocs/op
```

aggregator/handler/config.go
```diff
@@ -141,25 +144,31 @@ func (c *connectionConfiguration) NewConnectionOptions(scope tally.Scope) Connec
...
 
-func (c *partitionedHandlerConfiguration) NewHandler(
+func (c *shardedHandlerConfiguration) NewHandler(
+	totalShards int,
 	instrumentOpts instrument.Options,
 ) (aggregator.Handler, error) {
-	if err := c.validatePartitions(); err != nil {
-		return nil, err
+	if !c.DisableValidation {
+		if err := c.validateShards(totalShards); err != nil {
```
cw9: nit: handle `if !c.DisableValidation` in validateShards()?

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,116 @@
...
+	var (
+		// Divide overall queue size among all shards.
+		shardQueueSize = opts.QueueSize() / len(shards)
+		handlers       = make([]aggregator.Handler, totalShards)
+		instrumentOpts = opts.InstrumentOptions()
+		scope          = instrumentOpts.MetricsScope()
+	)
+	for _, shard := range shards {
+		handlerScope := scope.Tagged(map[string]string{
+			"shard-name": shard.Name,
```
cw9: nit: s/shard-name/shard-set-name? also rename BackendServerShardto BackendServerShardSet?

aggregator/types.go
```diff
@@ -33,14 +33,11 @@ import (
...
+// ShardFn maps a id to an aggregator shard given the total number of aggregator shards.
 type ShardFn func(id []byte, numShards int) uint32
 
-// PartitionFn maps a chunked id to a backend partition.
-type PartitionFn func(chunkedID id.ChunkedID) uint32
-
-// PartitionFnGen generates partition functions.
-type PartitionFnGen func() PartitionFn
+// AggregatedShardFn maps a chunked id of an aggregated metric to a shard.
+type AggregatedShardFn func(chunkedID id.ChunkedID, numShards int) uint32
```
cw9: Good name, should we rename ShardFn to UnaggregatedShardFn as well?

aggregator/handler/config.go
```diff
@@ -22,56 +22,72 @@ package handler
...
-	Type string `yaml:"type" validate:"regexp=(^blackhole$|^logging$|^forward$|^broadcast$)"`
+	Type Type `yaml:"type"`
 
 	// Forward handler configuration.
 	Forward *forwardHandlerConfiguration `yaml:"forward"`
 
+	// Sharded handler configuration.
+	Sharded *shardedHandlerConfiguration `yaml:"sharded"`
+
 	// Forward handler configuration.
```
jeromefroe: nit: should this be "Broadcast handler configuration"?

aggregator/handler/handler.go
```diff
@@ -20,13 +20,47 @@
...
 // Type is the handler type.
 type Type string
 
 // A list of supported handler types.
 const (
-	BlackholeHandler Type = "blackhole"
-	LoggingHandler   Type = "logging"
-	ForwardHandler   Type = "forward"
-	BroadcastHandler Type = "broadcast"
+	blackholeType Type = "blackhole"
```
jeromefroe: nit: any reason to not make these public? It seems a little odd that the configuration has a `Type` field, but none of these types are public although we unmarshal them correcly.

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,116 @@
...
+		shardNotAssigned: scope.Counter("shard-not-assigned"),
+	}
+}
+
+type shardedHandler struct {
+	handlers []aggregator.Handler
+	metrics  shardedHandlerMetrics
+}
+
+// NewShardedHandler creates a new sharded handler.
```
jeromefroe: Just for my own understanding, how do we handle updates to the shard topology, i.e. backends shards being added or removed? Seems it's just static and defined in configuration now, so I wanted to confirm.

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,116 @@
...
+var (
+	errEmptyBackendServerShards = errors.New("empty backend server shards")
+)
+
+// BackendServerShard contains a backend server shard.
+type BackendServerShard struct {
+	// Name of this shard for logging and metrics.
+	Name string `yaml:"name"`
+
+	// Shard range.
```
jeromefroe: nit: `Range` seems a little misleading here given that is in fact a set. Perhaps `Set` would be more appropriate and express the fact that it need not be a contiguous range?

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type shardedHandler struct {
+	handlers []aggregator.Handler
+	metrics  shardedHandlerMetrics
+}
+
+// NewShardedHandler creates a new sharded handler.
+func NewShardedHandler(
+	shards []BackendServerShard,
```
jeromefroe: nit: perhaps `shardSets` would be more appropriate than `shards`? Personally, when I see `shards` and `totalShards` my first thought is that `len(shards)` and `totalShards` should be equal.

services/m3aggregator/config/aggregator.go
```diff
@@ -531,50 +550,70 @@ func (c *kvClientConfiguration) NewKVClient(instrumentOpts instrument.Options) (
...
 }
 
-type shardFnType string
+type hashFnType string
 
-// List of supported sharding function types.
+// List of supported hashing function types.
 const (
-	murmur32ShardFn shardFnType = "murmur32"
+	murmur32HashFn hashFnType = "murmur32"
```
jeromefroe: nit: perhaps make this public as well since users will be expected to set in the configuration file?

services/m3aggregator/config/aggregator.go
```diff
@@ -531,50 +550,70 @@ func (c *kvClientConfiguration) NewKVClient(instrumentOpts instrument.Options) (
...
-		return nil, fmt.Errorf("unrecognized hash gen type %v", t)
+		return nil, fmt.Errorf("unrecognized shard function type %v", t)
+	}
+}
+
+func (t hashFnType) AggregatedShardFn() (aggregator.AggregatedShardFn, error) {
+	switch t {
+	case murmur32HashFn:
+		// NB(xichen): The function only allocates when the id of the aggregated metric
+		// is more than initialChunkedIDSize (i.e., 256 bytes) in size and requires zero
```
jeromefroe: super nit: seems `initialChunkedIDSize ` is actually set to 512 above.

aggregator/handler/common/queue.go
```diff
@@ -96,163 +104,163 @@ type forwardHandler struct {
...
-	defer h.wg.Done()
+func (q *queue) forwardToConn(addr string, conn *net.TCPConn) {
+	defer q.wg.Done()
 
 	var err error
 	continueFn := func(int) bool {
-		h.RLock()
-		closed := h.closed
-		h.RUnlock()
+		q.RLock()
```
jeromefroe: nit: might be more performant to use an [atomic `bool`](https://godoc.org/go.uber.org/atomic#Bool) here. We can replace the reads of `q.closed` with calls to `Load`, and in `Close` we can use `CAS`.

aggregator/handler/common/queue.go
```diff
@@ -96,163 +104,163 @@ type forwardHandler struct {
...
-			buf, ok := <-h.bufCh
+			buf, ok := <-q.bufCh
 			if !ok {
 				conn.Close()
-				atomic.AddInt32(&h.numOpenConnections, -1)
+				atomic.AddInt32(&q.numOpenConnections, -1)
 				return
 			}
-			_, writeErr := h.writeToConnFn(conn, buf.Buffer().Bytes())
+			_, writeErr := q.writeToConnFn(conn, buf.Buffer().Bytes())
```
jeromefroe: nit: perhaps just use `err` here?

aggregator/handler/common/queue.go
```diff
@@ -96,163 +104,163 @@ type forwardHandler struct {
...
+			q.log.WithFields(
 				log.NewField("address", addr),
 				log.NewErrField(writeErr),
 			).Error("error writing to server")
 
 			// NB(xichen): the buffer contains the oldest flushed data in queue
 			// so it's preferrable to drop it in case the queue is full.
-			if enqueueErr := h.enqueue(buf, dropCurrent); enqueueErr != nil {
-				h.log.WithFields(
+			if enqueueErr := q.enqueue(buf, dropCurrent); enqueueErr != nil {
```
jeromefroe: nit: same as above, maybe just `err`?

aggregator/handler/common/queue_test.go
```diff
@@ -39,49 +38,49 @@ const (
...
-	opts := testForwardHandlerOptions().SetQueueSize(3)
-	handler, err := newForwardHandler([]string{testFakeServerAddr}, opts)
+func TestQueueForwardToConnNoConnectionClosed(t *testing.T) {
+	opts := testQueueOptions().SetQueueSize(3)
+	q, err := newQueue([]string{testFakeServerAddr}, opts)
 	require.NoError(t, err)
-	// Queue up a nil buffer and close the handler.
-	handler.bufCh <- nil
-	handler.Close()
+	// Queue up a nil buffer and close the q.
```
jeromefroe: nit: s/the q/q

aggregator/handler/common/queue_test.go
```diff
@@ -210,61 +209,63 @@ func TestForwardHandlerForwardToConnReenqueueQueueFull(t *testing.T) {
...
-func TestForwardHandlerClose(t *testing.T) {
-	handler, err := newForwardHandler([]string{testFakeServerAddr}, testForwardHandlerOptions())
+func TestQueueClose(t *testing.T) {
+	q, err := newQueue([]string{testFakeServerAddr}, testQueueOptions())
 	require.NoError(t, err)
 
-	// Close the handler sets the flag.
-	handler.Close()
-	require.True(t, handler.closed)
+	// Close the q sets the flag.
```
jeromefroe: nit: s/the q/q

aggregator/handler/common/queue_test.go
```diff
@@ -210,61 +209,63 @@ func TestForwardHandlerForwardToConnReenqueueQueueFull(t *testing.T) {
...
+	q.Close()
+	require.True(t, q.closed)
 
-	// Close the handler a second time is a no op.
-	handler.Close()
-	require.True(t, handler.closed)
+	// Close the q a second time is a no op.
```
jeromefroe: nit: s/the q/q

aggregator/handler/common/router.go
```diff
@@ -0,0 +1,136 @@
...
+		routeNotFound: scope.Counter("route-not-found"),
+	}
+}
+
+type shardedRouter struct {
+	queues  []Queue
+	metrics shardedRouterMetrics
+}
+
+// RangedQueue is a backend queue responsible for a range of shards.
```
jeromefroe: nit: perhaps `ShardSetQueue` would be more appropriate? range gives me the impression that the shards must be contiguous which isn't the case

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,130 @@
...
+// THE SOFTWARE.
+
+package handler
+
+import (
+	"github.com/m3db/m3aggregator/aggregator"
+	"github.com/m3db/m3aggregator/aggregator/handler/common"
+	"github.com/m3db/m3aggregator/aggregator/handler/writer"
+	"github.com/m3db/m3aggregator/sharding"
+	"github.com/uber-go/tally"
```
jeromefroe: nit: newline between import?

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,130 @@
...
+	// Group routers by their sharder ids so that metrics are only encoded once
+	// for backends with the same sharding functions.
+	var (
+		nonShardedRouters  = make([]common.Router, 0, len(srs))
+		shardedRoutersByID = make([]sharderRouters, 0, len(srs))
+	)
+	for _, sr := range srs {
+		if sr.SharderID == sharding.NoShardingSharderID {
+			nonShardedRouters = append(nonShardedRouters, sr.Router)
+		} else {
```
jeromefroe: nit, take it or leave it: use `continue` so you don't need an `else` clause

aggregator/types.go
```diff
@@ -25,16 +25,15 @@ import (
...
 
 	"github.com/m3db/m3aggregator/aggregation/quantile/cm"
+	"github.com/m3db/m3aggregator/sharding"
+	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3metrics/policy"
-	"github.com/m3db/m3metrics/protocol/msgpack"
 	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3x/pool"
+	"github.com/uber-go/tally"
```
jeromefroe: nit: newline between imports?

jeromefroe:
Perhaps also add a README to the base of the aggregator directory capturing your comments on how all the interfaces fit together?

aggregator/handler/config.go
xichen2020: No longer applies

aggregator/handler/handler.go
xichen2020: Strictly speaking `Type` doesn't have to be public, I initially called it `handlerType` but considering it's inside the `handler` package, I got rid of the `handler` prefix.

aggregator/handler/config.go
xichen2020: Done

aggregator/types.go
xichen2020: Meh, I think the distinction is fairly clear and `UnaggregatedShardFn` is a handful to type.

aggregator/handler/sharded.go
xichen2020: Sounds good, done.

aggregator/handler/sharded.go
xichen2020: That's not supported at the moment, the backend topology is statically configured. That said at some point in the future we'll change the backend to read the topology from KV and route things dynamically based on KV updates to avoid deploying the agg tier when a backend server goes down.

aggregator/handler/sharded.go
xichen2020: No longer applies.

aggregator/handler/sharded.go
xichen2020: No longer applies.

services/m3aggregator/config/aggregator.go
xichen2020: This has been moved to the `sharding` package and yes the hashing types are now exported.

services/m3aggregator/config/aggregator.go
xichen2020: Updated.

aggregator/handler/common/queue.go
xichen2020: Sounds good.

aggregator/handler/common/queue.go
xichen2020: Yes the metrics have already been timestamped at this point. Sounds good will rename to `err`.

aggregator/handler/common/queue_test.go
xichen2020: Sounds good.

aggregator/handler/common/queue_test.go
xichen2020: Sounds good.

aggregator/handler/common/queue_test.go
xichen2020: Sounds good.

aggregator/handler/common/router.go
xichen2020: Renamed it to `ShardedQueue`.

aggregator/handler/sharded.go
xichen2020: Updated

aggregator/types.go
xichen2020: Sounds good.

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,131 @@
...
+	SharderID sharding.SharderID
+	Routers   []common.Router
+}
+
+type shardedHandler struct {
+	routersBySharderID []SharderRouter
+	writerOpts         writer.Options
+}
+
+func newShardedHandler(srs []SharderRouter, writerOpts writer.Options) aggregator.Handler {
```
jeromefroe: nit: is it worth making this constructor public as well since the other handler constructors are public, for example `NewLoggingHandler` and `NewBlackholeHandler`?

aggregator/handler/sharded.go
```diff
@@ -0,0 +1,131 @@
...
+			if !found {
+				shardedRoutersByID = append(shardedRoutersByID, sharderRouters{
+					SharderID: sr.SharderID,
+					Routers:   []common.Router{sr.Router},
+				})
+			}
+		}
+	}
+
+	// If there are both sharded routers and non-sharded routers, the non-sharded
```
jeromefroe: nit: do you think we'll have both non-sharded and sharded backends a lot in practice? This feels like a premature optimizations perhaps.

aggregator/handler/writer/multi.go
```diff
@@ -18,49 +18,49 @@
...
-type decodingHandler struct {
-	handle HandleFunc
+type multiWriter struct {
+	writers []aggregator.Writer
 }
 
-// NewDecodingHandler creates a new decoding handler with a custom handle function.
-func NewDecodingHandler(handle HandleFunc) aggregator.Handler {
-	return decodingHandler{handle: handle}
+// NewMultiWriter creates a new mulwriter.
```
jeromefroe: nit: `mulwriter`?

sharding/hash_test.go
```diff
@@ -0,0 +1,127 @@
...
+	shardFn, err := hashType.ShardFn()
+	require.NoError(t, err)
+
+	inputs := [][]byte{
+		[]byte("foo"),
+		[]byte("bar"),
+		[]byte("baz"),
+	}
+	for _, input := range inputs {
+		require.Equal(t, murmur3.Sum32(input)%uint32(numShards), shardFn(input, numShards))
```
jeromefroe: nit: instead of testing that it equals `murmur3.Sum32(input)%uint32(numShards)` perhaps we should test that it equals the actual value returned from that operation with the current package we are using? My concern is that we may bump the version of the package and that could unknowingly cause the output to change which may cause the collector and aggregator to get out of sync.

sharding/shard_set.go
```diff
@@ -0,0 +1,154 @@
...
+	"regexp"
+	"strconv"
+)
+
+const (
+	defaultNumShards = 1024
+)
+
+var (
+	// Shard range is expected to provided in the form of startShard..endShard.
```
jeromefroe: nit : s/expected to provided/expected to be provided/

sharding/shard_set.go
```diff
@@ -0,0 +1,154 @@
...
+	// Both startShard and endShard are inclusive. An example shard range is 0..63.
+	rangeRegexp = regexp.MustCompile(`^([0-9]+)(\.\.([0-9]+))?$`)
+
+	errInvalidShard = errors.New("invalid shard")
+)
+
+// ShardSet is a range of shards organized as a set.
```
jeromefroe: nit: perhaps its worth calling out that the set can be discontiguous?

sharding/shard_set.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// AddBetween adds shards between the given min (inclusive) and max (exclusive).
+func (ss ShardSet) AddBetween(minInclusive, maxExclusive uint32) {
+	for i := minInclusive; i < maxExclusive; i++ {
+		ss.Add(i)
+	}
+}
+
+// ParseRange parses a range of shards and adds them to the set.
```
jeromefroe: nit: perhaps `ParseAndAddRange` would be more appropriate?

aggregator/handler/common/queue_test.go
```diff
@@ -39,49 +38,49 @@ const (
...
+func TestQueueForwardToConnNoConnectionClosed(t *testing.T) {
+	opts := testQueueOptions().SetQueueSize(3)
+	q, err := newQueue([]string{testFakeServerAddr}, opts)
 	require.NoError(t, err)
-	// Queue up a nil buffer and close the handler.
-	handler.bufCh <- nil
-	handler.Close()
+	// Queue up a nil buffer and close the queue.
+	q.bufCh <- nil
+	q.Close()
```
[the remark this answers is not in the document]
xichen2020: Sounds good, can do it in a followup PR.

aggregator/handler/sharded.go
xichen2020: I'd actually make all of them private will do.

aggregator/handler/sharded.go
xichen2020: We have both sharded and non-sharded backends (ingester+indexer). Without this optimization we are paying the encoding the cost doubly for the non-sharded backends, which is on the critical path of flushes and can definitely negatively affect the performance if the flush latency gets out of control.

Given that the optimization only adds one extra line, I'd be willing to do that for the benefit of halving the cycles spent on encoding. This also enables users in the OSS world to add both sharded and non-sharded backends with relatively low cost.

aggregator/handler/writer/multi.go
xichen2020: Good catch, will fix.

sharding/hash_test.go
xichen2020: Sounds good will update.

sharding/shard_set.go
xichen2020: Sounds good.

sharding/shard_set.go
xichen2020: Sounds good.

sharding/shard_set.go
xichen2020: Hm feels a bit mouthful to do that, given that it's a member method seems self-explanatory to just use `ParseRange`.

aggregator/handler/common/router.go
```diff
@@ -0,0 +1,135 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package common
```
cw9: nit: move this to router/ package?

aggregator/handler/writer/sharded.go
```diff
@@ -0,0 +1,189 @@
...
+	// If the buffer size is not big enough, do nothing.
+	if sizeAfter < w.maxBufferSize {
+		return nil
+	}
+	// Otherwise we get a new buffer and copy the bytes exceeding the max
+	// flush size to it, swap the new buffer with the old one, and flush out
+	// the old buffer.
+	bufferedEncoder2 := w.bufferedEncoderPool.Get()
+	bufferedEncoder2.Reset()
+	if sizeBefore > 0 {
```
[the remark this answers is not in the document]
xichen2020: Can take a look in a followup PR. but afaik similar logic exists in the collector and the aggregator, in which case the collector uses unaggregated encoders and the aggregator uses aggregated encoders, so the common function will need to take a lambda of some sort due to lack of generics.

aggregator/list.go
```diff
@@ -438,7 +412,11 @@ func (l *metricLists) FindOrCreate(resolution time.Duration) (*metricList, error
 	}
 	list, exists = l.lists[resolution]
 	if !exists {
-		list = l.newMetricListFn(l.shard, resolution, l.opts)
+		var err error
+		list, err = l.newMetricListFn(l.shard, resolution, l.opts)
```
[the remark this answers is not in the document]
xichen2020: Meh feels more readable this way.

aggregator/types.go
```diff
@@ -98,16 +98,27 @@ type EntryPool interface {
 // QuantileSuffixFn returns the byte-slice suffix for a quantile value
 type QuantileSuffixFn func(quantile float64) []byte
 
-// Handler handles encoded streams containing aggregated metrics alongside their policies.
+// Handler handles aggregated metrics alongside their policies.
 type Handler interface {
```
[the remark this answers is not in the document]
xichen2020: The `Handler` hides the details of creating different types of writers (e.g., logging handler creates the logging writer, and the sharded handler creates the sharded writer, etc.). The user of the `Handler` interface has no knowledge of such details therefore is decoupled from the writers.

If we remove the `Handler` interface, such writer creation logic will need to be the responsibility of the users (the aggregator lists in this case), and for some of the writers (e.g., the sharded writer), it is not trivial. It seems to me having the `Handler` interface definitely provides better encapsulation.

aggregator/handler/common/router.go
xichen2020: I could but it's relatively simple (just `router.go` and `router_test.go`.

PR #53, opened 2017-09-25
Support including encoding time when writing aggregated metrics

xichen2020:
cc @jeromefroe @cw9 @prateek 

This PR adds logic to support including encoding time when writing aggregate metrics, which allows us to compute end-to-end latency metrics (e.g., prior to ingest).

### DIFF
--- a/aggregator/handler/config.go
+++ b/aggregator/handler/config.go
@@ -105,25 +105,18 @@ type writerConfiguration struct {
 	// Pool of buffered encoders.
 	BufferedEncoderPool pool.ObjectPoolConfiguration `yaml:"bufferedEncoderPool"`
 
-	// Whether to include encoding time in the payload.
-	IncludeEncodingTime *bool `yaml:"includeEncodingTime"`
-
 	// How frequent is the encoding time sampled and included in the payload.
-	EncodingTimeSamplingRate float64 `yaml:"encodingTimeSamplingRate"`
+	EncodingTimeSamplingRate float64 `yaml:"encodingTimeSamplingRate" validate:"min=0.0,max=1.0"`
 }
 
 func (c *writerConfiguration) NewWriterOptions(
 	instrumentOpts instrument.Options,
 ) writer.Options {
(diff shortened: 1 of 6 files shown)


aggregator/handler/writer/sharded.go
```diff
@@ -156,16 +164,36 @@ func (w *shardedWriter) encode(
...
 	buffer := bufferedEncoder.Buffer()
 	sizeBefore := buffer.Len()
-	if err := encoder.EncodeChunkedMetricWithStoragePolicy(mp); err != nil {
+
+	// Encode data.
+	var (
+		includeEncodingTime bool
+		err                 error
+	)
+	if w.includeEncodingTime {
```
jeromefroe: nit: would it perhaps be preferable to use `rand` to determine if we should include the encoding time? My concern is that if we sample every N metrics and there's some kind of bias in the latency of metrics that is dependent on the order of the metrics we may feel to detect it.

xichen2020: Sure can do.

aggregator/handler/config.go
```diff
@@ -113,6 +119,12 @@ func (c *writerConfiguration) NewWriterOptions(
 	if c.MaxBufferSize > 0 {
 		opts = opts.SetMaxBufferSize(c.MaxBufferSize)
 	}
+	if c.IncludeEncodingTime != nil {
```
jeromefroe: nit: any reason we wouldn't just set the sampling rate to zero if we didn't want to encode

xichen2020: In that case you'll just need to set `includeEncodingTime` to false.

jeromefroe: Yea, what I meant was perhaps it would be easier to get rid of this variable entirely.

xichen2020: Sure, was just trying to be explicit about it but yeah can do.

jeromefroe: I don't feel strongly, feel free to disregard.

PR #54, opened 2017-09-27
Remove deferred return and func for CPU savings

xichen2020:
cc @jeromefroe @cw9 

This PR removes deferred return and func execution in `Placement` on the hot path to save 2% CPU.


aggregator/placement_mgr.go
```diff
@@ -214,14 +214,14 @@ func (mgr *placementManager) placementWithLock() (placement.Placement, error) {
 		mgr.metrics.activeStagedPlacementErrors.Inc(1)
 		return nil, err
 	}
-	defer onStagedPlacementDoneFn()
```
jeromefroe: nit: maybe add a comment on why we don't use `defer` here?

xichen2020: Sounds good, will do

PR #58, opened 2017-09-29
Update m3cluster and m3metrics

cw9:
Use kv configuration from m3cluster
Decouple the election kv namespace with the kv config, so they don't have to be the same namespace(although they still can)

@xichen2020 @jeromefroe

### DIFF
--- a/services/m3aggregator/config/aggregator.go
+++ b/services/m3aggregator/config/aggregator.go
@@ -120,9 +120,6 @@ type AggregatorConfiguration struct {
 	// Client configuration for key value store.
 	KVClient kvClientConfiguration `yaml:"kvClient" validate:"nonzero"`
 
-	// Election KV namespace.
-	ElectionKVNamespace string `yaml:"electionKVNamespace" validate:"nonzero"`
-
 	// KV configuration.
 	KVConfig kv.Configuration `yaml:"kvConfig"`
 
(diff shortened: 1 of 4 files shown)


services/m3aggregator/config/aggregator.go
```diff
@@ -120,8 +120,11 @@ type AggregatorConfiguration struct {
 	// Client configuration for key value store.
 	KVClient kvClientConfiguration `yaml:"kvClient" validate:"nonzero"`
 
-	// KV namespace.
-	KVNamespace string `yaml:"kvNamespace" validate:"nonzero"`
+	// Election KV namespace.
+	ElectionKVNamespace string `yaml:"electionKVNamespace" validate:"nonzero"`
```
xichen2020: Remove this and simply use KVConfig.Namespace below?

cw9: Feels a little bit weird that way, although right now the namespace is the same for the placement and the election, but I feel they probably don't need to be tied together, also it's a bit weird that only namespace is used but not the environment. I could just use KVConfig.Namespace if you feel strongly.

xichen2020: Agreed that the namespaces don't have to be the same, however I'd prefer to have a single namespace in the config until there is a need to have two separate namespaces for placement and election.

cw9: Sounds good

cw9:
Sure, will do

PR #59, opened 2017-10-04
Less verbose logging and sampling error logs

xichen2020:
cc @cw9 @jeromefroe 

This PR removes some unnecessary verbose logging and sample error logs where necessary.

### DIFF
--- a/server/msgpack/server.go
+++ b/server/msgpack/server.go
@@ -22,6 +22,7 @@ package msgpack
 
 import (
 	"io"
+	"math/rand"
 	"net"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


server/msgpack/server.go
```diff
@@ -89,11 +96,15 @@ func (s *handler) Handle(conn net.Conn) {
...
 		policiesList := it.PoliciesList()
 		if err := s.aggregator.AddMetricWithPoliciesList(metric, policiesList); err != nil {
-			s.log.WithFields(
-				log.NewField("metric", metric.String()),
-				log.NewField("policies", policiesList),
-				log.NewErrField(err),
-			).Errorf("error adding metric with policies")
+			// We sample the error log here because the error rate may scale with
+			// the metrics incoming rate and consume lots of cpu cycles.
+			if s.rand.Float64() < s.errLogSamplingRate {
```
jeromefroe: We might want to look at `zap` at some point, we would get this for free with them.

xichen2020: Yeah, I think there's an open PR in m3x that sets the zap logger on the instrument options but the last update was 20 days ago. I'd be happy to use it when it's ready but this will get me unblocked.

PR #60, opened 2017-10-09
Refactor kv configuration to support different logging levels for different KV stores

xichen2020:
cc @cw9 @jeromefroe 

This PR refactors the kv configuration for the agg tier to support different logging levels for different KV stores. Because m3cluster currently logs every kv update, and the agg tier persists flush times information to KV every 10s by default, this pollutes the log files and makes it difficult to find real issues. This PR moves kv config to individual manager configuration so the logging levels for different stores can be configured differently (e.g., use a log level of `warn` for the store that watches and persists flush times to reduce verbosity of logging)

### DIFF
--- a/config/m3aggregator.yaml
+++ b/config/m3aggregator.yaml
@@ -101,10 +101,10 @@ aggregator:
           capacity: 32
         - count: 1024
           capacity: 64
-  kvConfig:
-    namespace: /r2/m3aggregator
-    environment: production
   placementManager:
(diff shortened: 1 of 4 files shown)


xichen2020:
Are you saying we should have a customizable logging level per watch? That seems a bit excessive and probably requires API breaking changes (e.g., `store.Watch()` would require passing in a desired logger etc).

PR #61, opened 2017-10-12
Update quantile suffix function

cw9:
Updated the quantile function to generate ".p9999" rather than ".p99.99" for P9999

@xichen2020

### DIFF
--- a/aggregator/options.go
+++ b/aggregator/options.go
@@ -21,7 +21,6 @@
 package aggregator
 
 import (
-	"fmt"
 	"strconv"
 	"strings"
 	"sync"
@@ -797,11 +796,7 @@ func (o *options) computeDefaultSuffixes() {
 		default:
 			q, ok := aggType.Quantile()
 			if ok {
-				b, err := o.timerQuantileSuffixFn(q)
-				if err != nil {
-					panic(fmt.Sprintf("could not generate default suffix for %f: %v", q, err))
-				}
-				o.defaultSuffixes[aggType.ID()] = b
+				o.defaultSuffixes[aggType.ID()] = o.timerQuantileSuffixFn(q)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


aggregator/options.go
```diff
@@ -796,7 +797,11 @@ func (o *options) computeDefaultSuffixes() {
 		default:
 			q, ok := aggType.Quantile()
 			if ok {
-				o.defaultSuffixes[aggType.ID()] = o.timerQuantileSuffixFn(q)
+				b, err := o.timerQuantileSuffixFn(q)
```
xichen2020: Perhaps it'll be better to still have the function return a `[]byte` and panic inside the function on error instead?

aggregator/options.go
```diff
@@ -869,8 +874,20 @@ func (o *options) computeFullGaugePrefix() {
...
 // By default we use e.g. ".p50", ".p95", ".p99" for the 50th/95th/99th percentile.
-func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	return []byte(".p" + strconv.FormatFloat(quantile*100, 'f', -1, 64))
+func defaultTimerQuantileSuffixFn(quantile float64) ([]byte, error) {
+	if quantile < 0 || quantile > 1 {
+		return nil, fmt.Errorf("invalid quantile %f", quantile)
+	}
+	str := strconv.FormatFloat(quantile*100, 'f', -1, 64)
+
+	if strings.Contains(str, ".") {
```
xichen2020: nit: Can directly call `strings.Index(str, ".")` here and check it against -1.

aggregator/options.go
```diff
@@ -869,8 +874,20 @@ func (o *options) computeFullGaugePrefix() {
 }
 
 // By default we use e.g. ".p50", ".p95", ".p99" for the 50th/95th/99th percentile.
-func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	return []byte(".p" + strconv.FormatFloat(quantile*100, 'f', -1, 64))
+func defaultTimerQuantileSuffixFn(quantile float64) ([]byte, error) {
+	if quantile < 0 || quantile > 1 {
```
xichen2020: Probably don't need this? Alternatively this should be validated when setting the quantiles?

aggregator/options.go
```diff
@@ -869,8 +874,20 @@ func (o *options) computeFullGaugePrefix() {
...
+		return nil, fmt.Errorf("invalid quantile %f", quantile)
+	}
+	str := strconv.FormatFloat(quantile*100, 'f', -1, 64)
+
+	if strings.Contains(str, ".") {
+		if quantile <= 0.99 {
```
xichen2020: Perhaps we just always remove the dot regardless?

cw9: Yeah, I'm ok either way, gonna remove the error/panic and just remove the dot, we have control over the valid aggregation types list anyway.

aggregator/options.go
cw9: sg

aggregator/options.go
```diff
@@ -870,7 +870,13 @@ func (o *options) computeFullGaugePrefix() {
 
 // By default we use e.g. ".p50", ".p95", ".p99" for the 50th/95th/99th percentile.
 func defaultTimerQuantileSuffixFn(quantile float64) []byte {
-	return []byte(".p" + strconv.FormatFloat(quantile*100, 'f', -1, 64))
+	str := strconv.FormatFloat(quantile*100, 'f', -1, 64)
+
```
xichen2020: nit: remove the superfluous blank line.

PR #62, opened 2017-10-16
Use aggregation types config from m3metrics

cw9:


aggregator/options.go
```diff
@@ -215,38 +131,14 @@ func NewOptions() Options {
 	return o
 }
 
-func (o *options) SetDefaultCounterAggregationTypes(aggTypes policy.AggregationTypes) Options {
+func (o *options) SetAggregationTypesOptions(value policy.AggregationTypesOptions) Options {
```
xichen2020: nit: perhaps the aggregation types options should be grouped with other options (e.g., clock options etc)?

glide.yaml
```diff
@@ -6,11 +6,11 @@ import:
 - package: github.com/m3db/m3cluster
   version: b9650bbb69955b5daf276cd840bd36cc986e9dc9
 - package: github.com/m3db/m3metrics
-  version: 82d18cdd2a24502c6a170b6bfea01c30685e5b26
+  version: bb6198075758dd454c3f0d2b0bb3f640162394ba
 - package: google.golang.org/appengine
   version: 2e4a801b39fc199db615bfca7d0b9f8cd9580599
 - package: github.com/m3db/m3x
-  version: 118fa2f66efa708eb1afdcd756c4da0ec6a8848e
+  version: 5120e5bcfaf4888c15052e5c9d7ba700dce4247c
```
xichen2020: nit: might as well update to b009e41c873b0fb0fc066114911c87ff6a6f8d77?

services/m3aggregator/config/aggregator.go
```diff
@@ -776,20 +655,12 @@ func (b jitterBucketsByIntervalAscending) Less(i, j int) bool {
 	return b[i].FlushInterval < b[j].FlushInterval
 }
 
-// timerQuantileSuffixFnType is the timer quantile suffix function type.
```
xichen2020: Do we no longer allow users to customize the quantile suffix functions by providing a different type?

aggregator/options.go
cw9: sounds good

glide.yaml
cw9: sure

services/m3aggregator/config/aggregator.go
cw9: We don't have another suffix type for user to config to anyway, we need a code change to add the alternative suffix func we could add the config in that diff, also the config should be in m3metric.

xichen2020: Sure sounds good.

PR #65, opened 2017-10-19
Do not update cached policies when receiving tombstoned policies

xichen2020:
cc @cw9 @jeromefroe 

This PR fix an issue when ingesting metrics with tombstoned policies. Previously if the policies indicate the metric is tombstoned, the internal policies are updated and cached, and the entry is marked as tombstoned. This causes an issue when there are two different raw metrics matching the same rollup rule, one of which matches the rule and the other does not. As a result, the agg tier will receive a tombstoned policy for the second raw metric, mark the entry tombstoned, even though the entry is still very much alive due to the first raw metric still matching that rule and generating datapoints for that entry.

The fix is simply to not update the internal cached policies when receiving a tombstoned policy and instead rely on background ticking goroutine to sweep entries that have not been recently accessed.

### DIFF
--- a/aggregator/entry_test.go
+++ b/aggregator/entry_test.go
@@ -93,7 +93,6 @@ func TestEntryResetSetData(t *testing.T) {
 	require.False(t, e.hasDefaultPoliciesList)
 	require.False(t, e.useDefaultPolicies)
 	require.Equal(t, int64(uninitializedCutoverNanos), e.cutoverNanos)
-	require.False(t, e.tombstoned)
 	require.Equal(t, lists, e.lists)
 	require.Equal(t, int32(0), e.numWriters)
 	require.Equal(t, now.UnixNano(), e.lastAccessNanos)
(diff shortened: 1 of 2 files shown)


aggregator/entry_test.go
```diff
@@ -807,9 +802,13 @@ func testEntryAddMetricWithPoliciesList(
 				id := elem.Value.(*CounterElem).ID()
 				require.Equal(t, testCounterID, id)
 				aggregations := elem.Value.(*CounterElem).values
-				require.Equal(t, 1, len(aggregations))
-				require.Equal(t, alignedStart.UnixNano(), aggregations[0].timeNanos)
-				require.Equal(t, int64(1234), aggregations[0].counter.Sum())
+				if !expectedShouldAdd {
```
cw9: nit: 
`if expectedShouldAdd {...} else {...}`

aggregator/entry_test.go
```diff
@@ -818,9 +817,13 @@ func testEntryAddMetricWithPoliciesList(
 				id := elem.Value.(*TimerElem).ID()
 				require.Equal(t, testBatchTimerID, id)
 				aggregations := elem.Value.(*TimerElem).values
-				require.Equal(t, 1, len(aggregations))
-				require.Equal(t, alignedStart.UnixNano(), aggregations[0].timeNanos)
-				require.Equal(t, 18.0, aggregations[0].timer.Sum())
+				if !expectedShouldAdd {
```
cw9: same

aggregator/entry_test.go
xichen2020: Meh I usually put the shorter block first to make it easier on the eyes.

aggregator/entry_test.go
xichen2020: Same as above.

PR #66, opened 2017-10-22
Retain flush times seen for KV persistence

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to retain flush times seen for kv persistence to fix a race condition during a topology change in which the leader instance closes a leaving shard and updates the persisted flush times before the follower instance gets a chance to process the persisted flush times in order to close the leaving shard.

The PR also refactored the follower flush manager and distinguishes between received flush times and process times, the latter of which is used to determine whether the follower instance can take over leadership, which is safer than using received times for that purpose.

### DIFF
--- a/aggregator/follower_flush_mgr.go
+++ b/aggregator/follower_flush_mgr.go
@@ -75,7 +75,8 @@ type followerFlushManager struct {
 	scope                 tally.Scope
 
 	doneCh          <-chan struct{}
-	proto           *schema.ShardSetFlushTimes
+	received        *schema.ShardSetFlushTimes
+	processed       *schema.ShardSetFlushTimes
 	flushTimesState flushTimesState
 	flushMode       followerFlushMode
 	lastFlushed     time.Time
@@ -140,6 +141,7 @@ func (mgr *followerFlushManager) Prepare(buckets []*flushBucket) (flushTask, tim
 	)
 	if mgr.flushTimesState == flushTimesUpdated {
 		mgr.flushTimesState = flushTimesProcessed
+		mgr.processed = mgr.received
 		mgr.flushMode = kvUpdateFollowerFlush
 		flushersByInterval = mgr.flushersFromKVUpdateWithLock(buckets)
 		needsFlush = true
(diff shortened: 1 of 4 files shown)


aggregator/leader_flush_mgr.go
```diff
@@ -215,26 +222,41 @@ func (mgr *leaderFlushManager) computeNextFlushNanos(flushInterval time.Duration
...
+			mgr.flushedNanos[flusherID] = flusher.LastFlushedNanos()
+		}
+	}
+}
+
 func (mgr *leaderFlushManager) prepareFlushTimesWithLock(
 	buckets []*flushBucket,
 ) schema.ShardSetFlushTimes {
+	// Collect the flush times from all the flushers in the buckets.
+	mgr.collectFlushTimesWithLock(buckets)
```
jeromefroe: nit: it seems a little odd to me that `flushedNanos` is a field on the `leaderFlushManager` even though it is only used once, on line 249. Perhaps it would clearer for `collectFlushTimesWithLock ` to return the map explicitly?

xichen2020: The `flushedNanos` field is to retain all the flushed times that the leader has ever seen, including the flush times of shards that have been closed. This is so that during a topology change, if the leader closes out a shard, the flush times of that shard will still be persisted in kv for the follower to receive so the follower can flush and close the shard accordingly. 

Otherwise, there is a race condition during a topology change where the leader closes a leaving shard and flushes, and the flush times persisted in kv will no longer include the latest flush times of the closed shard, which means the follower won't ever close that leaving shard due to missing flush times of that shard.

jeromefroe: Oh, I see, so `Prepare` and therefore `collectFlushTimesWithLock ` will be called repeatedly?

xichen2020: Correct, `Prepare` is called every second by default by the flush manager (https://github.com/m3db/m3aggregator/blob/master/aggregator/flush_mgr.go#L282)

aggregator/follower_flush_mgr.go
```diff
@@ -75,7 +71,8 @@ type followerFlushManager struct {
 	scope                 tally.Scope
 
 	doneCh          <-chan struct{}
-	proto           *schema.ShardSetFlushTimes
+	received        *schema.ShardSetFlushTimes
```
cw9: add a comment about why we have 2 different proto here?

PR #69, opened 2017-11-14
Rate limiting per-metric writes

xichen2020:
cc @cw9 @jeromefroe 

This PR adds rate limiting logic to limit the write rate per metric. The limit is backed by KV and disabled by default. This protects aggregation tier from being overwhelmed with high frequency metric emissions, and in extreme cases will allow us to perform load shedding to selectively drop values high-frequency metrics when the aggregation tier performance is suffering.

### DIFF
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -496,6 +496,11 @@ func (e *Entry) resetRateLimiterWithLock(runtimeOpts runtime.Options) {
 		e.rateLimiter = nil
 		return
 	}
+	if e.rateLimiter == nil {
+		nowFn := e.opts.ClockOptions().NowFn()
+		e.rateLimiter = rate.NewLimiter(newLimit, nowFn)
+		return
+	}
 	e.rateLimiter.Reset(newLimit)
 }
 
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -28,6 +28,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/m3db/m3aggregator/rate"
+	"github.com/m3db/m3aggregator/runtime"
 	metricid "github.com/m3db/m3metrics/metric/id"
 	"github.com/m3db/m3metrics/metric/unaggregated"
 	"github.com/m3db/m3metrics/policy"
@@ -37,12 +39,15 @@ import (
 )
 
 var (
-	errEmptyPoliciesList = errors.New("empty policies list")
-	errEntryClosed       = errors.New("entry is closed")
+	errEmptyPoliciesList      = errors.New("empty policies list")
+	errEntryClosed            = errors.New("entry is closed")
+	errEntryRateLimitExceeded = errors.New("entry rate limit is exceeded")
 )
 
 type entryMetrics struct {
 	emptyPoliciesList tally.Counter
+	rateLimitExceeded tally.Counter
+	droppedValues     tally.Counter
 	stalePolicy       tally.Counter
 	futurePolicy      tally.Counter
 	tombstonedPolicy  tally.Counter
@@ -52,6 +57,8 @@ type entryMetrics struct {
 func newEntryMetrics(scope tally.Scope) entryMetrics {
 	return entryMetrics{
 		emptyPoliciesList: scope.Counter("empty-policies-list"),
+		rateLimitExceeded: scope.Counter("rate-limit-exceeded"),
+		droppedValues:     scope.Counter("dropped-values"),
 		stalePolicy:       scope.Counter("stale-policy"),
 		futurePolicy:      scope.Counter("future-policy"),
 		tombstonedPolicy:  scope.Counter("tombstoned-policy"),
@@ -65,6 +72,7 @@ type Entry struct {
 
 	closed                 bool
 	opts                   Options
+	rateLimiter            *rate.Limiter
 	hasDefaultPoliciesList bool
 	useDefaultPolicies     bool
 	cutoverNanos           int64
@@ -79,14 +87,14 @@ type Entry struct {
 }
 
 // NewEntry creates a new entry.
-func NewEntry(lists *metricLists, opts Options) *Entry {
+func NewEntry(lists *metricLists, runtimeOpts runtime.Options, opts Options) *Entry {
 	scope := opts.InstrumentOptions().MetricsScope().SubScope("entry")
 	e := &Entry{
 		aggregations: make(map[policy.Policy]*list.Element),
 		metrics:      newEntryMetrics(scope),
 		decompressor: policy.NewPooledAggregationIDDecompressor(opts.AggregationTypesOptions().AggregationTypesPool()),
 	}
-	e.ResetSetData(lists, opts)
+	e.ResetSetData(lists, runtimeOpts, opts)
 	return e
 }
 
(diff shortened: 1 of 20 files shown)


services/m3aggregator/config/aggregator.go
```diff
@@ -377,13 +391,107 @@ type kvClientConfiguration struct {
...
+) (client.Client, error) {
 	if c.Etcd == nil {
 		return nil, errNoKVClientConfiguration
 	}
 	return c.Etcd.NewClient(instrumentOpts)
 }
 
+type runtimeOptionsConfiguration struct {
+	KVConfig                              kv.Configuration `yaml:"kvConfig"`
+	WriteValuesPerMetricLimitPerSecondKey string           `yaml:"writeValuesPerMetricLimitPerSecondKey" validate:"nonzero"`
```
xichen2020: Should remove the nonzero validation.

services/m3aggregator/config/aggregator.go
```diff
@@ -377,13 +391,107 @@ type kvClientConfiguration struct {
...
+	store kv.Store,
+	limitKey string,
+	defaultLimit int64,
+	runtimeOptsManager aggruntime.OptionsManager,
+	logger log.Logger,
+) {
+	limit := defaultLimit
+	value, err := store.Get(limitKey)
+	if err == nil {
+		var protoValue commonpb.Int64Proto
```
cw9: use kvutil.Int64FromValue?

services/m3aggregator/config/aggregator.go
```diff
@@ -377,13 +391,107 @@ type kvClientConfiguration struct {
...
+
+	watch, err := store.Watch(limitKey)
+	if err != nil {
+		logger.Errorf("unable to watch per-metric write value limit: %v", err)
+		return
+	}
+	go func() {
+		for range watch.C() {
+			value := watch.Get()
+			if value == nil {
```
cw9: use kvutil.Int64FromValue?

services/m3aggregator/config/aggregator.go
```diff
@@ -303,7 +317,7 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	entryPoolOpts := c.EntryPool.NewObjectPoolOptions(iOpts)
 	entryPool := aggregator.NewEntryPool(entryPoolOpts)
 	opts = opts.SetEntryPool(entryPool)
-	entryPool.Init(func() *aggregator.Entry { return aggregator.NewEntry(nil, opts) })
+	entryPool.Init(func() *aggregator.Entry { return aggregator.NewEntry(nil, initRuntimeOpts, opts) })
```
cw9: pass in runtimeOptsManager.RuntimeOptions() here instead of initRuntimeOpts?

aggregator/map.go
```diff
@@ -123,88 +145,62 @@ func (m *metricMap) Tick(target time.Duration) tickResult {
...
-		// Actually purging the expired entries.
-		if len(expired) >= batchSize {
-			numExpired += m.purgeExpired(now, expired)
-			for i := range expired {
-				expired[i] = emptyHashedEntry
-			}
-			expired = expired[:0]
-		}
+	// NB(xichen): we hold onto the entry list deletion lock here to ensure no
+	// entries get deleted while we iterate over the list, otherwise we may update
```
cw9: What happens if we update the expired entries? aren't they gonna be deleted later anyway? 
It would be much better to not having to have a delete lock here, using the 2 locks adds complexity and make the code prone to dead locks.

config/m3aggregator.yaml
```diff
@@ -92,6 +92,12 @@ aggregator:
           capacity: 32
         - count: 1024
           capacity: 64
+  runtimeOptions:
+    kvConfig:
+      namespace: /m3aggregator
+      environment: production
+    writeValuesPerMetricLimitPerSecondKey: write-values-per-metric-limit-per-second
+    WriteValuesPerMetricLimitPerSecond: 0
```
cw9: lower case w

runtime/options_manager.go
```diff
@@ -0,0 +1,97 @@
...
+func (w *optionsManager) RegisterWatcher(
+	watcher OptionsWatcher,
+) close.SimpleCloser {
+	_, watch, _ := w.watchable.Watch()
+
+	// The watchable is always initialized so it's okay to do a blocking read.
+	<-watch.C()
+
+	// Deliver the current runtime options.
+	watcher.SetRuntimeOptions(watch.Get().(Options))
```
cw9: Is it safe to directly do `watcher.SetRuntimeOptions(w.watchable.RuntimeOptions())` without waiting on <-watch.C()?

aggregator/entry.go
```diff
@@ -460,3 +489,32 @@ func (e *Entry) shouldExpire(now time.Time) bool {
...
+	e.RLock()
+	rateLimiter := e.rateLimiter
+	e.RUnlock()
+	if rateLimiter == nil {
+		return nil
+	}
+	if rateLimiter.IsAllowed(numValues) {
+		return nil
+	}
+	e.metrics.rateLimitExceeded.Inc(1)
```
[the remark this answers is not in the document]
xichen2020: This metric just records how many times rate limit has been exceeded. The `droppedValues` metric below records how many values have been dropped.

aggregator/map.go
xichen2020: I'd argue deleting expired entries is more dangerous. Expired entries are put back into the object pool (or reused), so we'd be examining / mutating state of an object we should not even have access to because it's already expired.

There are only 2 places where the deletion lock is used, and as long as the locks are acquired in the same order, we can ensure no deadlocks are going to happen. This is primarily needed for the sake of performance so that entries can be inserted while the runtime options are updated.

config/m3aggregator.yaml
xichen2020: Ah good catch, will do, thanks

rate/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+// Limit returns the current limit.
+func (l *Limiter) Limit() int64 {
+	l.RLock()
+	limit := l.limitPerSecond
+	l.RUnlock()
+	return limit
+}
+
+// IsAllowed returns whether n events may happen now.
+func (l *Limiter) IsAllowed(n int64) bool {
```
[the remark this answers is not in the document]
xichen2020: `IsAllowed` feels more natural to me, `Allow` is a verb that technically should return an error.

rate/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	alignedNow := l.nowFn().Truncate(time.Second)
+	l.RLock()
+	if !alignedNow.After(l.alignedLast) {
+		isAllowed := atomic.AddInt64(&l.allowed, n) <= l.limitPerSecond
```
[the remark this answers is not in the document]
xichen2020: Sure, sounds good.

rate/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+
+	l.Lock()
+	if !alignedNow.After(l.alignedLast) {
+		isAllowed := atomic.AddInt64(&l.allowed, n) <= l.limitPerSecond
+		l.Unlock()
+		return isAllowed
+	}
+	l.alignedLast = alignedNow
+	l.allowed = n
+	isAllowed := l.allowed <= l.limitPerSecond
```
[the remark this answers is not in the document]
xichen2020: That means we still allow requests after a previous request is rejected, which is probably what we want.

rate/limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	l.Unlock()
+	return isAllowed
+}
+
+// Reset resets the internal state.
+func (l *Limiter) Reset(limit int64) {
```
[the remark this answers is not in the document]
xichen2020: In theory yes, in practice `int` and `int64` are used far more widely than `uint64`, even the golang rate limiter uses `int`.

runtime/options_manager.go
xichen2020: We could, but if we have a watch here, might as well use it.

services/m3aggregator/config/aggregator.go
xichen2020: Sure sounds good.

services/m3aggregator/config/aggregator.go
xichen2020: Sure sounds good.

services/m3aggregator/config/aggregator.go
xichen2020: Sure sounds good.

aggregator/map.go
cw9: Can this be replace by doing a entry.ShouldExpire check before SetRuntimeOptions()?

xichen2020: The issue is we don't know when an entry is expired or not without the lock, the entry being examined could be expired from the ticking goroutine from underneath us, and we'd be calling `ShouldExpire` on the entry that's already expired.

PR #70, opened 2017-11-14
Rate limit error logging

xichen2020:
cc @cw9 @jeromefroe 

This PR adds rate limiting for error logs, so that for cases where we rate limit writes, the error logs are limited to a fix number of logs / second, instead of the full set.

### DIFF
--- a/server/msgpack/server.go
+++ b/server/msgpack/server.go
@@ -103,6 +103,7 @@ func (s *handler) Handle(conn net.Conn) {
 		metric := it.Metric()
 		policiesList := it.PoliciesList()
 		if err := s.aggregator.AddMetricWithPoliciesList(metric, policiesList); err != nil {
+			s.metrics.addMetricErrors.Inc(1)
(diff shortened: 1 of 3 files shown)


server/msgpack/server.go
```diff
@@ -96,17 +103,18 @@ func (s *handler) Handle(conn net.Conn) {
...
+				s.metrics.errLogRateLimited.Inc(1)
+				continue
 			}
+			s.log.WithFields(
+				log.NewField("type", metric.Type.String()),
+				log.NewField("id", metric.ID.String()),
+				log.NewField("policies", policiesList),
+				log.NewErrField(err),
+			).Errorf("error adding metric with policies")
 			s.metrics.addMetricErrors.Inc(1)
```
jeromefroe: nit: should we increment this even if we don't log the error? Otherwise we have to sum this and `errLogRateLimited ` which may be unintuitive.

xichen2020: Right yeah makes sense.

PR #71, opened 2017-11-17
Rate limiting new metric insertions after warmup

xichen2020:
cc @cw9 @jeromefroe 

This PR adds rate limiting for new metric insertions after a configurable warmup period so that if a user started sending a large amount of high-cardinality metrics within a short time, the new entry creation requests will be rate limited to protect the aggregation tier from being overwhelmed.

### DIFF
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -39,30 +39,30 @@ import (
 )
 
 var (
-	errEmptyPoliciesList      = errors.New("empty policies list")
-	errEntryClosed            = errors.New("entry is closed")
-	errEntryRateLimitExceeded = errors.New("entry rate limit is exceeded")
+	errEmptyPoliciesList           = errors.New("empty policies list")
+	errEntryClosed                 = errors.New("entry is closed")
+	errWriteValueRateLimitExceeded = errors.New("write value rate limit is exceeded")
 )
 
 type entryMetrics struct {
-	emptyPoliciesList tally.Counter
-	rateLimitExceeded tally.Counter
-	droppedValues     tally.Counter
-	stalePolicy       tally.Counter
-	futurePolicy      tally.Counter
-	tombstonedPolicy  tally.Counter
-	policyUpdates     tally.Counter
+	emptyPoliciesList      tally.Counter
+	valueRateLimitExceeded tally.Counter
+	droppedValues          tally.Counter
+	stalePolicy            tally.Counter
+	futurePolicy           tally.Counter
+	tombstonedPolicy       tally.Counter
+	policyUpdates          tally.Counter
 }
 
 func newEntryMetrics(scope tally.Scope) entryMetrics {
 	return entryMetrics{
-		emptyPoliciesList: scope.Counter("empty-policies-list"),
-		rateLimitExceeded: scope.Counter("rate-limit-exceeded"),
-		droppedValues:     scope.Counter("dropped-values"),
-		stalePolicy:       scope.Counter("stale-policy"),
-		futurePolicy:      scope.Counter("future-policy"),
-		tombstonedPolicy:  scope.Counter("tombstoned-policy"),
-		policyUpdates:     scope.Counter("policy-updates"),
+		emptyPoliciesList:      scope.Counter("empty-policies-list"),
+		valueRateLimitExceeded: scope.Counter("value-rate-limit-exceeded"),
+		droppedValues:          scope.Counter("dropped-values"),
+		stalePolicy:            scope.Counter("stale-policy"),
+		futurePolicy:           scope.Counter("future-policy"),
+		tombstonedPolicy:       scope.Counter("tombstoned-policy"),
+		policyUpdates:          scope.Counter("policy-updates"),
 	}
 }
 
@@ -140,7 +140,7 @@ func (e *Entry) AddMetricWithPoliciesList(
 	switch mu.Type {
 	case unaggregated.BatchTimerType:
 		var err error
-		if err = e.applyRateLimit(int64(len(mu.BatchTimerVal))); err == nil {
+		if err = e.applyValueRateLimit(int64(len(mu.BatchTimerVal))); err == nil {
 			err = e.writeBatchTimerWithPoliciesList(mu, pl)
 		}
 		if mu.BatchTimerVal != nil && mu.TimerValPool != nil {
@@ -149,7 +149,7 @@ func (e *Entry) AddMetricWithPoliciesList(
 		return err
 	default:
 		// For counters and gauges, there is a single value in the metric union.
-		if err := e.applyRateLimit(1); err != nil {
+		if err := e.applyValueRateLimit(1); err != nil {
 			return err
 		}
 		return e.addMetricWithPoliciesList(mu, pl)
@@ -504,7 +504,7 @@ func (e *Entry) resetRateLimiterWithLock(runtimeOpts runtime.Options) {
 	e.rateLimiter.Reset(newLimit)
 }
 
-func (e *Entry) applyRateLimit(numValues int64) error {
+func (e *Entry) applyValueRateLimit(numValues int64) error {
 	e.RLock()
 	rateLimiter := e.rateLimiter
 	e.RUnlock()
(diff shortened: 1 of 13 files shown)


config/m3aggregator.yaml
```diff
@@ -57,6 +57,16 @@ msgpack:
     iteratorPool:
       size: 4096
 
+runtimeOptions:
+  kvConfig:
+    namespace: /m3aggregator
+    environment: production
+  writeValuesPerMetricLimitPerSecondKey: write-values-per-metric-limit-per-second
+  writeValuesPerMetricLimitPerSecond: 0
+  writeNewMetricLimitClusterPerSecondKey: write-new-metric-limit-per-second
```
cw9: nit: can this be new-metric(-limit)-per-shard-per-second?

config/m3aggregator.yaml
```diff
@@ -57,6 +57,16 @@ msgpack:
     iteratorPool:
       size: 4096
 
+runtimeOptions:
+  kvConfig:
+    namespace: /m3aggregator
+    environment: production
+  writeValuesPerMetricLimitPerSecondKey: write-values-per-metric-limit-per-second
```
cw9: nit: can this be write-frequency( or qps)(-limit)-per-metric-per-second?

services/m3aggregator/config/runtime.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	// If watch creation failed for both, we return immediately.
+	if valueLimitCh == nil && newMetricLimitCh == nil {
+		return
+	}
+
+	go func() {
+		for {
+			select {
+			case <-valueLimitCh:
```
cw9: if valueLimitCh == nil && newMetricLimitCh != nil, we could be pulling from a nil valueLimitCh here?

services/m3aggregator/config/runtime.go
```diff
@@ -0,0 +1,189 @@
...
+		valueLimitCh = valueLimitWatch.C()
+	}
+	newMetricLimitWatch, err := store.Watch(newMetricClusterLimitKey)
+	if err != nil {
+		logger.Errorf("unable to watch cluster-wide write new metric limit: %v", err)
+	} else {
+		newMetricLimitCh = newMetricLimitWatch.C()
+	}
+	// If watch creation failed for both, we return immediately.
+	if valueLimitCh == nil && newMetricLimitCh == nil {
```
cw9: should this check be ||?

services/m3aggregator/config/runtime.go
```diff
@@ -0,0 +1,189 @@
...
+				}
+				logger.Infof("updating per-shard write new metric limit from %d to %d", currNewMetricPerShardLimit, newNewMetricPerShardLimit)
+				runtimeOpts = runtimeOpts.SetWriteNewMetricLimitPerShardPerSecond(newNewMetricPerShardLimit)
+				runtimeOptsManager.SetRuntimeOptions(runtimeOpts)
+			}
+		}
+	}()
+}
+
+func clusterLimitToPerShardLimit(
```
cw9: Is it easier to just set the limit per shard rather than per cluster?

aggregator/entry.go
```diff
@@ -39,30 +39,30 @@ import (
 )
 
 var (
-	errEmptyPoliciesList      = errors.New("empty policies list")
-	errEntryClosed            = errors.New("entry is closed")
-	errEntryRateLimitExceeded = errors.New("entry rate limit is exceeded")
+	errEmptyPoliciesList           = errors.New("empty policies list")
+	errEntryClosed                 = errors.New("entry is closed")
+	errWriteValueRateLimitExceeded = errors.New("write value rate limit is exceeded")
```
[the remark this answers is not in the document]
xichen2020: It meant "the rate at which we are writing values exceeded rate limit". The other rate limit error is similarly named as `errWriteNewMetricRateLimit...`.

aggregator/entry.go
```diff
@@ -140,7 +140,7 @@ func (e *Entry) AddMetricWithPoliciesList(
 	switch mu.Type {
 	case unaggregated.BatchTimerType:
 		var err error
-		if err = e.applyRateLimit(int64(len(mu.BatchTimerVal))); err == nil {
+		if err = e.applyValueRateLimit(int64(len(mu.BatchTimerVal))); err == nil {
```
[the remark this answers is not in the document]
xichen2020: It identifies the thing that we are applying rate limiting to. Feels like `frequency` is already captured by the "rate" in "rate limit", no?

aggregator/map.go
```diff
@@ -321,4 +351,35 @@ func (m *metricMap) forEachEntry(entryFn hashedEntryFn) {
...
+		return
+	}
+	m.rateLimiter.Reset(newLimit)
+}
+
+func (m *metricMap) applyNewMetricRateLimitWithLock(now time.Time) error {
+	if m.rateLimiter == nil {
+		return nil
+	}
+	noLimitWarmupDuration := m.runtimeOpts.WriteNewMetricNoLimitWarmupDuration()
```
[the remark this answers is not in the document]
xichen2020: Sure. It'll be in two places, though.

aggregator/map.go
```diff
@@ -321,4 +351,35 @@ func (m *metricMap) forEachEntry(entryFn hashedEntryFn) {
...
+	if warmupEnd := m.firstInsertAt.Add(noLimitWarmupDuration); now.Before(warmupEnd) {
+		m.metrics.noRateLimitWarmup.Inc(1)
+		return nil
+	}
+	if m.rateLimiter.IsAllowed(1) {
+		return nil
+	}
+	m.metrics.newMetricRateLimitExceeded.Inc(1)
+	m.metrics.droppedNewMetrics.Inc(1)
```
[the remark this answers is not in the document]
xichen2020: Not sure what you mean? These are emitted when the object being rate limited is dropped (i.e., the rate liimit is exceeded). We do the same for value rate limiting, except that in that case we may drop a batch of values when the rate limit is exceeded.

config/m3aggregator.yaml
xichen2020: But we are writing metric values, no? As mentioned above, the idea of "frequency" is already captured by "value per metric per second"?

config/m3aggregator.yaml
xichen2020: This is the cluster-wide limit per second, not the per-shard limit. Otherwise whoever sets this needs to do math to figure out what the proper limit is for the entire cluster.

services/m3aggregator/config/runtime.go
xichen2020: No, if we can't watch one, at least we can watch the other.

services/m3aggregator/config/runtime.go
xichen2020: Yes, reading from a nil channel blocks forever.

services/m3aggregator/config/runtime.go
xichen2020: It's easier to write code that way, but harder to reason about for the person who needs to set it. Say you want to set the limit, you can say I want to limit the new metric rate for the agg tier to be capped at 1M/s, and you just set it. However, if the limit is per shard, you'll need to figure out how many shards there are, what the replication factor is, and do math to figure that out, which is worse operationally.

aggregator/entry.go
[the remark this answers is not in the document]
cw9: value feels like the value of the metric rather than the length of the timer/metric batch though, can it just be errWriteRateLimitExceeded and the new metrics one be errNewMetricRateLimit...?

aggregator/map.go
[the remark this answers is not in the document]
cw9: I felt odd that the 2 metrics will always have the exact value but if this is the convention to match the other case then it's fine

services/m3aggregator/config/runtime.go
cw9: I thought this value is mainly to rate limit the map insertion in each agg box? if it's set per cluster, then you need to use x/8k*num_shard_per_instance to know how many insertions is happening on each m3agg box?

PR #79, opened 2017-12-04
Resign from ongoing campaign when election manager is closed

### LINKED ISSUE #72: Aggregator should stop campaigning against old lock when its shard set id is updated
cc @cw9 

Currently when the shard set id of a node is updated, the aggregator will reset its shard set id internally but not stop campaigning against the old lock key in M3KV. This is normally a non-issue since each node should be uniquely associated with a shard set id and shard set ids are not reused. However, in the case we do reuse them (e.g., during development), this causes an issue in which the old nodes were still holding on to the old lock, causing the new nodes owning the same shard set id unable to acquire the lock until the old nodes are restarted.

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to asynchronously resign from the ongoing campaign when the election manager is closed to avoid having lingering campaigns during process shutdown or shard set id reset.

Fixes #72 .

### DIFF
--- a/aggregator/election_mgr.go
+++ b/aggregator/election_mgr.go
@@ -187,6 +187,9 @@ type electionManagerMetrics struct {
 	followerResign                         tally.Counter
 	resignTimeout                          tally.Counter
 	resignErrors                           tally.Counter
+	resignOnCloseSuccess                   tally.Counter
+	resignOnCloseErrors                    tally.Counter
+	resignOnClose                          tally.Gauge
 	followerToPendingFollower              tally.Counter
 	electionState                          tally.Gauge
 	campaignState                          tally.Gauge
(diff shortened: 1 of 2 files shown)


aggregator/election_mgr.go
```diff
@@ -728,6 +736,19 @@ func (mgr *electionManager) campaignLoop(campaignStateWatch watch.Watch) {
 			}
 			mgr.processCampaignUpdate(campaignStatus)
 		case <-mgr.doneCh:
+			electionKey := mgr.electionKey
+			// Asynchronously resign from ongoing campaign on close to avoid blocking the close
```
jeromefroe: Why do we want to avoid blocking the `Close` call? This seems a little to odd me because now the `Close` call can return without ensuring all the resources used by the `electionManager` are actually closed.

xichen2020: Talked offline, `Close` still waits for all goroutines to exit before the call completes. It resigns asynchronously because it may take a long while to resign and this may be called while aggregator lock is held, which we don't want to happen. We simply need to ensure the ongoing campaign will be cancelled but it doesn't matter whether it's done synchronously or asynchronously.

PR #84, opened 2017-12-14
Distinguish different types of errors for better visibility

### LINKED ISSUE #83: Distinguish different types of errors for better alerting
cc @cw9 

We are currently not distinguishing error types and as such alerting on a single metric. So if a non-critical error happens (e.g., some metric is being rate limited), this triggers a high-urgency alert, which is not great. Distinguishing the types of errors would allow us to configure the alerts more effectively.

xichen2020:
cc @cw9 @jeromefroe 

This PR distinguishes different types of errors reported in the aggregator for better visibility and easier alerting configurations.

Fixes #83

### DIFF
--- a/aggregator/aggregator_test.go
+++ b/aggregator/aggregator_test.go
@@ -574,7 +574,7 @@ func TestAggregatorAddMetricMetrics(t *testing.T) {
 	// Validate we count successes and errors correctly.
 	require.Equal(t, 4, len(counters))
 	for _, id := range []string{
-		"testScope.success+", // NB: this looks like a bug with tally test scope
+		"testScope.success+",
 		"testScope.errors+reason=value-rate-limit-exceeded",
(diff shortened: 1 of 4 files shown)


aggregator/aggregator_test.go
```diff
@@ -560,6 +560,45 @@ func TestAggregatorOwnedShards(t *testing.T) {
...
+		c, exists := counters[id]
+		require.True(t, exists)
+		require.Equal(t, int64(1), c.Value())
+	}
+
+	// Validate we record times correctly.
+	require.Equal(t, 1, len(timers))
+
+	for _, id := range []string{
+		"testScope.success-latency+", // NB: this looks like a bug with tally test scope
```
jeromefroe: We talked about this in the past: https://github.com/uber-go/tally/pull/41 I think the decision was to not break backwards compatibility.

PR #89, opened 2018-03-13
Use auto-generated mocks

xichen2020:
cc @jeromefroe @cw9 

This PR replaces hand-written mocks with auto-generated mocks for the sake of simplicity when writing unit tests.


glide.yaml
```diff
@@ -69,8 +69,11 @@ import:
 - package: github.com/ghodss/yaml
   version: 73d445a93680fa1a78ae23a5839bad48f32ba1ee
 
+# NB(xichen): need to pull in a change to support returning
```
prateek: lol, bleeding edge :)

aggregator/aggregator_mock.go
```diff
@@ -0,0 +1,467 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/m3db/m3aggregator/generated/proto/flush"
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3cluster/shard"
+	"github.com/m3db/m3metrics/metric/aggregated"
+	"github.com/m3db/m3x/watch"
+	"github.com/uber-go/tally"
```
prateek: looks like the `mockclean` utility is failing on the input file here (it should generate output with the correct import blocks). Mind opening an issue on `github.com/m3db/build-tools` to track

xichen2020: Sure yeah can do

xichen2020: Btw I put up a PR to fix this issue: https://github.com/m3db/ci-scripts/pull/35. It's not a mockclean issue per se. Rather, it's because the auto-generated files are not finding imports in the right location. In any case, feel free to take a look.

PR #90, opened 2018-03-16
 Update m3cluster and m3metrics

cw9:


aggregator/aggregator_mock.go
```diff
@@ -18,340 +18,396 @@
...
 	"context"
+	"reflect"
 	"time"
 
 	"github.com/m3db/m3aggregator/generated/proto/flush"
 	"github.com/m3db/m3cluster/placement"
 	"github.com/m3db/m3cluster/shard"
 	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3x/watch"
+	"github.com/uber-go/tally"
```
xichen2020: Hey so this is not ordered properly due to an issue I fixed in the latest rev of ci-scripts, can you pull that in and re-run mockgen? Thanks.

integration/custom_aggregation_test.go
```diff
@@ -26,16 +26,17 @@ import (
...
 
 	"github.com/stretchr/testify/require"
 )
 
 var (
-	compressor                             = policy.NewAggregationIDCompressor()
-	compressedMin, _                       = compressor.Compress(policy.AggregationTypes{policy.Min})
-	compressedMinAndMax, _                 = compressor.Compress(policy.AggregationTypes{policy.Min, policy.Max})
+	compressor                             = aggregation.NewIDCompressor()
+	compressedMin, _                       = compressor.Compress(aggregation.Types{aggregation.Min})
```
xichen2020: Can you change this and the following one to use `MustCompress`? Thanks

services/m3aggregator/config/aggregator.go
```diff
@@ -369,7 +371,7 @@ func (c *streamConfiguration) NewStreamOptions(instrumentOpts instrument.Options
 }
 
 type placementManagerConfiguration struct {
-	KVConfig         kv.Configuration               `yaml:"kvConfig"`
+	KVConfig         kv.OverrideConfiguration       `yaml:"kvConfig"`
```
xichen2020: Do we need to set loggers here as well?

aggregator/aggregator_mock.go
cw9: sounds good

integration/custom_aggregation_test.go
cw9: sounds good

services/m3aggregator/config/aggregator.go
cw9: Nope, I think normal logger here is good enough, for flush times mgr we need to config to warn level logger

PR #91, opened 2018-03-19
Refactor interfaces and move them into their own files and packages

xichen2020:
cc @cw9 @jeromefroe 

This PR refactors the interfaces and move them into their own files and packages for more clear separation of responsibilities.


aggregator/types.go
```diff
@@ -1,308 +0,0 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
```
cw9: any particular reasons to not keep the types.go pattern?

aggregator/handler/handler.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2018 Uber Technologies, Inc.
```
cw9: rename to types.go?

aggregator/types.go
xichen2020: I feel it's easier to look at the interfaces alongside their implementations, unless there are multiple implementations of the interface, in which case it makes sense to pull out the interface into its own file. Don't think it's a conventional pattern to put all interfaces in one large `types.go` file even though they sometimes aren't closely related.

aggregator/handler/handler.go
xichen2020: There's already a `type.go` file defining different handler types hence the `handler.go`.

PR #92, opened 2018-03-23
Ingest untimed metrics with pipeline metadata

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to ingest untimed metrics with pipeline metadata at the aggregation tier.

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -242,7 +242,7 @@ type aggregatorMetrics struct {
 }
 
 func newAggregatorMetrics(scope tally.Scope, samplingRate float64) aggregatorMetrics {
-	addMetricScope := scope.SubScope("addUntimed")
+	addUntimedScope := scope.SubScope("addUntimed")
 	placementScope := scope.SubScope("placement")
 	shardsScope := scope.SubScope("shards")
 	shardSetIDScope := scope.SubScope("shard-set-id")
@@ -252,7 +252,7 @@ func newAggregatorMetrics(scope tally.Scope, samplingRate float64) aggregatorMet
 		timers:       scope.Counter("timers"),
 		timerBatches: scope.Counter("timer-batches"),
 		gauges:       scope.Counter("gauges"),
-		addUntimed:   newAggregatorAddUntimedMetrics(addMetricScope, samplingRate),
+		addUntimed:   newAggregatorAddUntimedMetrics(addUntimedScope, samplingRate),
 		placement:    newAggregatorPlacementMetrics(placementScope),
 		shards:       newAggregatorShardsMetrics(shardsScope),
 		shardSetID:   newAggregatorShardSetIDMetrics(shardSetIDScope),
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -32,9 +32,10 @@ import (
 	"github.com/m3db/m3aggregator/sharding"
 	"github.com/m3db/m3cluster/placement"
 	"github.com/m3db/m3cluster/shard"
+	"github.com/m3db/m3metrics/metadata"
+	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3metrics/metric/id"
 	"github.com/m3db/m3metrics/metric/unaggregated"
-	"github.com/m3db/m3metrics/policy"
 	"github.com/m3db/m3x/clock"
 	"github.com/m3db/m3x/instrument"
 
@@ -46,8 +47,8 @@ type Aggregator interface {
 	// Open opens the aggregator.
 	Open() error
 
-	// AddMetricWithPoliciesList adds a metric with policies list for aggregation.
-	AddMetricWithPoliciesList(mu unaggregated.MetricUnion, pl policy.PoliciesList) error
+	// AddUntimed adds an untimed metric with staged metadatas.
+	AddUntimed(metric unaggregated.MetricUnion, metas metadata.StagedMetadatas) error
 
 	// Resign stops the aggregator from participating in leader election and resigns
 	// from ongoing campaign if any.
@@ -58,6 +59,12 @@ type Aggregator interface {
 
 	// Close closes the aggregator.
 	Close() error
+}
+
+// metricsAggregator contains private aggregator APIs.
+type metricsAggregator interface {
+	// addForwarded adds a forwarded metric with metadata.
+	addForwarded(metric aggregated.Metric, metadata metadata.ForwardMetadata) error
 }
 
 const (
@@ -72,7 +79,7 @@ var (
 	errShardNotOwned                 = errors.New("aggregator shard is not owned")
 )
 
-type aggregatorAddMetricMetrics struct {
+type aggregatorAddUntimedMetrics struct {
 	success                    tally.Counter
 	successLatency             tally.Timer
 	invalidMetricTypes         tally.Counter
@@ -83,11 +90,11 @@ type aggregatorAddMetricMetrics struct {
 	uncategorizedErrors        tally.Counter
 }
 
-func newAggregatorAddMetricMetrics(
+func newAggregatorAddUntimedMetrics(
 	scope tally.Scope,
 	samplingRate float64,
-) aggregatorAddMetricMetrics {
-	return aggregatorAddMetricMetrics{
+) aggregatorAddUntimedMetrics {
+	return aggregatorAddUntimedMetrics{
 		success:        scope.Counter("success"),
 		successLatency: instrument.MustCreateSampledTimer(scope.Timer("success-latency"), samplingRate),
 		invalidMetricTypes: scope.Tagged(map[string]string{
@@ -111,12 +118,12 @@ func newAggregatorAddMetricMetrics(
 	}
 }
 
-func (m *aggregatorAddMetricMetrics) ReportSuccess(d time.Duration) {
+func (m *aggregatorAddUntimedMetrics) ReportSuccess(d time.Duration) {
 	m.success.Inc(1)
 	m.successLatency.Record(d)
 }
 
-func (m *aggregatorAddMetricMetrics) ReportError(err error) {
+func (m *aggregatorAddUntimedMetrics) ReportError(err error) {
 	if err == nil {
 		return
 	}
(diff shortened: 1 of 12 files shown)


aggregator/entry.go
```diff
@@ -28,60 +28,76 @@ import (
...
+		valueRateLimitExceeded:  scope.Counter("value-rate-limit-exceeded"),
+		droppedValues:           scope.Counter("dropped-values"),
+		staleMetadata:           scope.Counter("stale-metadata"),
+		tombstonedMetadata:      scope.Counter("tombstoned-metadata"),
+		metadataUpdates:         scope.Counter("metadata-updates"),
 	}
 }
 
-// Entry stores metadata about current policies and aggregations for a metric.
+// Entry keeps track of all aggregations of a metric alongside their aggregation
```
cw9: nit: 'their' -> 'its'

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
 	}
-	if useDefaultPolicies {
-		policies = e.opts.DefaultPolicies()
-	}
-	return e.hasPolicyChangesWithLock(policies)
-}
 
-func (e *Entry) hasPolicyChangesWithLock(newPolicies []policy.Policy) bool {
-	if len(e.aggregations) != len(newPolicies) {
+	// If this is a newer metadata, we always update.
```
cw9: Do we still need to check future metadata? Seems to be in the old logic?

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
-	for _, p := range policies {
-		if elem, exists := e.aggregations[p]; exists {
-			newAggregations[p] = elem
-		} else {
-			aggTypes, err := e.decompressor.Decompress(p.AggregationID)
-			if err != nil {
-				return err
-			}
+	// Otherwise it is necessary to make a copy because it's not owned by us.
+	elemID := make(metricid.RawID, len(metric.ID))
```
cw9: should this be pooled?

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
+					}
+					newElem = e.opts.GaugeElemPool().Get()
+				default:
+					return errInvalidMetricType
 				}
-				newElem = e.opts.TimerElemPool().Get()
-			case unaggregated.GaugeType:
-				if !aggTypes.IsValidForGauge() {
-					return fmt.Errorf("invalid aggregation types %s for Gauge", aggTypes.String())
+				// NB: The pipeline may not be owned by us and as such we need to make a copy here.
```
cw9: who else might own the pipeline?

aggregator/entry.go
```diff
@@ -28,60 +28,76 @@ import (
...
 	"github.com/m3db/m3metrics/metric/unaggregated"
+	"github.com/m3db/m3metrics/op/applied"
 	"github.com/m3db/m3metrics/policy"
 	xerrors "github.com/m3db/m3x/errors"
 
 	"github.com/uber-go/tally"
 )
 
+const (
+	initialAggregationCapacity = 2
```
cw9: why use 2 here?

aggregator/entry.go
```diff
@@ -211,36 +226,44 @@ func (e *Entry) addMetricWithPoliciesList(
...
 		timeLock.RUnlock()
-		e.metrics.tombstonedPolicy.Inc(1)
+		e.metrics.tombstonedMetadata.Inc(1)
 		return nil
 	}
 
-	if !e.shouldUpdatePoliciesWithLock(currTime, sp) {
-		err := e.addMetricWithLock(currTime, mu)
+	// It is expected that there is at least one pipeline in the metadata.
+	if len(sm.Pipelines) == 0 {
```
cw9: Does this mean all the metrics need to carry a pipeline over the network?

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
-	// Fast path to exit in case the policies didn't change.
-	if !e.hasPolicyChangesWithLock(policies) {
-		e.hasDefaultPoliciesList = hasDefaultPoliciesList
-		e.useDefaultPolicies = useDefaultPolicies
-		e.cutoverNanos = sp.CutoverNanos
-		e.metrics.policyUpdates.Inc(1)
-		return nil
+func (e *Entry) maybeCopyIDWithLock(metric unaggregated.MetricUnion) metricid.RawID {
+	// If we own the ID, there is no need to copy.
+	if metric.OwnsID {
```
cw9: when does the metric not own its ID?

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
+}
+
+func (e *Entry) updateMetadatasWithLock(
+	metric unaggregated.MetricUnion,
+	hasDefaultMetadatas bool,
+	sm metadata.StagedMetadata,
+) error {
+	var (
+		elemID          = e.maybeCopyIDWithLock(metric)
+		newAggregations = make(aggregationValues, 0, initialAggregationCapacity)
```
cw9: Not sure if it's necessary but since this is always locked, I think you can get away with keeping 2 aggregations in the entry: existingAggregations and newAggregations, when you need to update, do it to the newAggregations and swap the 2 in the end of the update.

aggregator/aggregator.go
```diff
@@ -60,6 +61,12 @@ type Aggregator interface {
 	Close() error
 }
 
+// metricsAggregator contains private aggregator APIs.
+type metricsAggregator interface {
+	// addForwarded adds a forwarded metric with metadata.
+	addForwarded(metric aggregated.Metric, metadata metadata.ForwardMetadata) error
```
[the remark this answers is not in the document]
xichen2020: Sure I can make the method public.

aggregator/entry.go
xichen2020: Mainly just because most metrics will have `10s:2d` and `1m:40d` by default, so the initial capacity of 2.

aggregator/entry.go
xichen2020: Sounds good.

aggregator/entry.go
xichen2020: Yes. Although in most cases it'll be an empty pipeline and the encoder can optimize for that case, but either way it won't be a significant portion of the network usage.

aggregator/entry.go
xichen2020: So right now the iterator always reuses the ID during decoding, because the iterator is per-connection, it can reuse ID across iterations as long as the ID is copied after the iterator moves onto the next iteration.

aggregator/entry.go
xichen2020: I don't think that's necessary at the moment, this should only happen if it's a brand new ID we have never seen before, which is very rare.

aggregator/entry.go
xichen2020: True, but given that this only happens once after rule update and should be infrequent, seems like an overkill to make that optimization now.

aggregator/entry.go
xichen2020: So the thinking here is that the pipeline might be reused by the decoding iterator, just like it does with metric IDs, and as such we need to make a copy to ensure the pipeline doesn't get changed underneath us.

aggregator/entry.go
xichen2020: So I changed the logic in `activeStagedMetadataWithLock` to return an error if we can't find a staged metadata that has been cut over, and as a result, the metadata passed to `shouldUpdateMetadatasWithLock` should always have been cut over (see the NB above `shouldUpdateMetadatasWithLock`).

aggregator/entry.go
```diff
@@ -296,10 +319,11 @@ func (e *Entry) TryExpire(now time.Time) bool {
 		return false
 	}
 	e.closed = true
-	for p, agg := range e.aggregations {
-		agg.Value.(metricElem).MarkAsTombstoned()
-		delete(e.aggregations, p)
+	for i := range e.aggregations {
+		e.aggregations[i].elem.Value.(metricElem).MarkAsTombstoned()
+		e.aggregations[i] = aggregationValue{}
```
jeromefroe: How come we're resetting the the entries here when we truncate the slice at the end of the loop? Is this to release any internal pointers?

aggregator/entry.go
xichen2020: Sure thing.

aggregator/entry.go
xichen2020: Correct, even if you reset the slice, the backing array still holds pointers to internal fields, so this is commonly done to avoid GC sweeping overhead.

xichen2020: Sure will do.

aggregator/aggregator.go
```diff
@@ -223,19 +230,19 @@ func newAggregatorShardSetIDMetrics(scope tally.Scope) aggregatorShardSetIDMetri
...
+	addUntimed   aggregatorAddUntimedMetrics
+	placement    aggregatorPlacementMetrics
+	shards       aggregatorShardsMetrics
+	shardSetID   aggregatorShardSetIDMetrics
+	tick         aggregatorTickMetrics
 }
 
 func newAggregatorMetrics(scope tally.Scope, samplingRate float64) aggregatorMetrics {
-	addMetricScope := scope.SubScope("addMetricWithPoliciesList")
+	addMetricScope := scope.SubScope("addUntimed")
```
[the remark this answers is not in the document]
xichen2020: Yeah good catch, will do ta

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
 	}
-	for _, policy := range newPolicies {
-		if _, exists := e.aggregations[policy]; !exists {
-			return true
+
+	// Iterate over the list of pipelines and check whether we have metadata changes.
+	// NB: If the incoming metadata have the same set of aggregation keys as the cached
+	// set but also have duplicates, there is no need to update metadatas as long as
+	// the cached set has all aggregation keys in the incoming metadata and vice versa.
+	bs := bitset.New(uint(len(e.aggregations)))
```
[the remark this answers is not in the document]
xichen2020: The idea here is that the bitset will be allocated on the stack as long as the number of unique pipelines is under 64, and as such there's no need to reuse it.

aggregator/entry.go
```diff
@@ -315,170 +339,179 @@ func (e *Entry) recordLastAccessed(currTime time.Time) {
...
-			}
-		} else {
-			// Otherwise this is a new id so it is necessary to make a
-			// copy because it's not owned by us.
-			elemID = make(metricid.RawID, len(id))
-			copy(elemID, id)
+	// If there are existing elements for this id, try reusing
+	// the id from the elements because those are owned by us.
+	if len(e.aggregations) > 0 {
+		for _, aggVal := range e.aggregations {
```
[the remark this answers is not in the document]
xichen2020: Hm yeah not sure why it was done that way, will change ta

PR #96, opened 2018-03-30
Aggregation elements flush or forward metrics based on pipeline metadata

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to the aggregation elements to process aggregated metrics by flushing or forwarding them based on pipeline metadata.

NB: the changes in `*.gen.go` are auto-generated from `generic_elem.go`.

### DIFF
--- a/aggregator/elem_base.go
+++ b/aggregator/elem_base.go
@@ -42,7 +42,10 @@ const (
 	defaultNumValues = 2
 
 	// Maximum transformation derivative order that is supported.
-	maxSupportedTransformationDerivativeOrer = 1
+	// A default value of 1 means we currently only support transformations that
+	// compute first-order derivatives. This applies to the most common usecases
+	// without imposing signficant bookkeeping overhead.
+	maxSupportedTransformationDerivativeOrder = 1
 )
 
 var (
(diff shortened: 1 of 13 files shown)


aggregator/elem_base.go
```diff
@@ -37,19 +40,15 @@ import (
 const (
 	// Default number of aggregation buckets allocated initially.
 	defaultNumValues = 2
-)
 
-var (
-	errElemClosed = errors.New("element is closed")
+	// Maximum transformation derivative order that is supported.
+	maxSupportedTransformationDerivativeOrer = 1
```
cw9: typo: ...Derivative`Order`

aggregator/elem_base.go
```diff
@@ -239,5 +260,89 @@ func (e gaugeElemBase) NewLockedAggregation(_ Options, aggOpts raggregation.Opti
...
+			// A binary transformation is a transformation that computes first-order derivatives.
+			if transformOp.Type.IsBinaryTransform() {
+				transformationDerivativeOrder++
+			}
+		}
+	}
+	if firstRollupOpIdx == -1 {
+		return parsedPipeline{}, fmt.Errorf("pipeline %v has no rollup operations", pipeline)
+	}
+	if transformationDerivativeOrder > maxSupportedTransformationDerivativeOrer {
```
cw9: can you elaborate what does this mean? could you provide a pipeline example that will error out in this case?

services/m3aggregator/config/aggregator.go
```diff
@@ -263,19 +263,20 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
...
 
-	// Set default policies.
-	policies := make([]policy.Policy, len(c.DefaultPolicies))
+	// Set default storage policies.
+	// TODO(xichen): sort the storage policies.
+	policies := make([]policy.StoragePolicy, len(c.DefaultPolicies))
 	copy(policies, c.DefaultPolicies)
-	sort.Sort(policy.ByResolutionAscRetentionDesc(policies))
-	opts = opts.SetDefaultPolicies(policies)
+	// sort.Sort(policy.ByResolutionAscRetentionDesc(policies))
```
cw9: still need this?

aggregator/elem_base.go
```diff
@@ -37,19 +40,15 @@ import (
 const (
 	// Default number of aggregation buckets allocated initially.
 	defaultNumValues = 2
-)
 
-var (
-	errElemClosed = errors.New("element is closed")
+	// Maximum transformation derivative order that is supported.
```
cw9: Also comment why use 1 here?

aggregator/elem_base.go
xichen2020: Good catch, ta.

aggregator/elem_base.go
xichen2020: Talked offline, so pipelines that compute higher order derivatives require keeping more states including the raw values and lower order derivatives, and as such are not supported by the current implementation. An example is "aggregate Last | perSecond | perSecond". Will add comments to clarify.

services/m3aggregator/config/aggregator.go
xichen2020: Yeah I'll leave it as is, will clean it up when I address the TODOs.

aggregator/elem_base.go
xichen2020: Sure yeah.

PR #97, opened 2018-04-05
Better test coverage for new aggregator logic

xichen2020:
cc @cw9 @jeromefroe 

This PR fixes the existing tests and adds new tests for the aggregator logic.

NB: The integration tests and the server tests are currently disabled since the encoding/decoding API for metric metadata has not been implemented yet.

### DIFF
--- a/aggregator/capture/aggregator.go
+++ b/aggregator/capture/aggregator.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 
 	aggr "github.com/m3db/m3aggregator/aggregator"
+	"github.com/m3db/m3metrics/metadata"
 	"github.com/m3db/m3metrics/metric/id"
 	"github.com/m3db/m3metrics/metric/unaggregated"
 	"github.com/m3db/m3metrics/policy"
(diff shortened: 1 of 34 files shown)


aggregator/capture/aggregator.go
```diff
@@ -49,43 +50,40 @@ func NewAggregator() Aggregator {
...
+func (agg *aggregator) AddUntimed(
 	mu unaggregated.MetricUnion,
-	pl policy.PoliciesList,
+	sm metadata.StagedMetadatas,
 ) error {
-	// Clone the metric and policies to ensure it cannot be mutated externally.
+	// Clone the metric and metadatas to ensure it cannot be mutated externally.
 	mu = cloneMetric(mu)
-	pl = clonePoliciesList(pl)
+	sm = cloneStagedMetadatas(sm)
```
cw9: Do we always need to clone this? If so could we just return new object from iterator? Or maybe the caller of AddUntimed should do the clone before sending the objects in?

aggregator/generic_elem.go
```diff
@@ -360,24 +361,29 @@ func (e *GenericElem) processValue(
 				prev := transformation.Datapoint{TimeNanos: e.lastConsumedAtNanos, Value: e.lastConsumedValues[aggTypeIdx]}
 				curr := transformation.Datapoint{TimeNanos: timeNanos, Value: value}
 				res := fn(prev, curr)
-				value = res.Value
 				// NB: we only need to record the value needed for derivative transformations.
 				// We currently only support first-order derivative transformations so we only
 				// need to keep one value. In the future if we need to support higher-order
 				// derivative transformations, we need to store an array of values here.
 				e.lastConsumedValues[aggTypeIdx] = value
+				value = res.Value
```
cw9: Is there a test for this?

aggregator/capture/aggregator.go
xichen2020: Yea the iterator might reuse staged metadatas to reduce allocation overhead (otherwise we might need to allocate a lot more for the common case), hence the need for cloning. The "capturing" aggregator is only for testing purposes so it's okay to always clone. The real aggregator only clones when the metadata changes which is a lot more efficient.

PR #101, opened 2018-04-22
Add aggregation server client

xichen2020:
cc @cw9 @jeromefroe 

This PR adds the client implementation for writing metrics to the aggregation server. Most of the implementation were moved from `backend/server` package in `m3collector`, with the msgpack encoder replaced with protobuf encoder and hand-written mocks replaced with auto-generated mocks.


client/writer.go
```diff
@@ -0,0 +1,376 @@
...
+
+			// Otherwise we enqueue the current buffer.
+			buffer := w.prepareEnqueueBufferWithLock(encoder, sizeBefore)
+
+			// Unlock the encoder before we enqueue the old buffer to ensure other
+			// goroutines have an oppurtunity to encode metrics while larger timer
+			// batches are being encoded.
+			encoder.Unlock()
+
+			enqueueErr = w.enqueueBuffer(buffer)
```
cw9: the encoder lock usage here is unconventional and kinda hard to maintain, how much time does this enqueueBuffer call take? is it really necessary to do it outside of lock?

client/writer.go
```diff
@@ -0,0 +1,376 @@
...
+
+			// Re-lock the encoder and update variables since the encoder's buffer
+			// may have been updated.
+			encoder.Lock()
+			sizeBefore = encoder.Len()
+
+			if enqueueErr != nil {
```
cw9: handle enqueueErr here instead of handle it later outside the loop?

client/writer.go
```diff
@@ -0,0 +1,376 @@
...
+				break
+			}
+
+			// If the buffer isn't big enough continue to the next iteration.
+			if sizeAfter := encoder.Len(); sizeAfter < w.flushSize {
+				continue
+			}
+
+			// Otherwise we enqueue the current buffer.
+			buffer := w.prepareEnqueueBufferWithLock(encoder, sizeBefore)
```
cw9: why is this function called for timers specifically than reuse the logic outside the loop?

client/writer.go
```diff
@@ -0,0 +1,376 @@
...
+	for _, encoder := range w.encodersByShard {
+		encoder.Lock()
+		if encoder.Len() == 0 {
+			encoder.Unlock()
+			continue
+		}
+		buffer := encoder.Relinquish()
+		encoder.Unlock()
+		if err := w.enqueueBuffer(buffer); err != nil {
+			multiErr = multiErr.Add(err)
```
cw9: does user handle the err from just one encoder?

client/client.go
```diff
@@ -0,0 +1,233 @@
...
+		if !ok {
+			err = fmt.Errorf("instance %s does not own shard %d", instance.ID(), shardID)
+			multiErr = multiErr.Add(err)
+			continue
+		}
+		if !c.shouldWriteForShard(nowNanos, shard) {
+			continue
+		}
+		if err = c.writerMgr.WriteUntimed(instance, shardID, metric, metadatas); err != nil {
+			multiErr = multiErr.Add(err)
```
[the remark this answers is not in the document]
xichen2020: They can log the error and emit metrics to track the errors. Retrying on error is probably not a good idea as the writes are not idempotent. The only time when the error could be surfaced is that the instance queue is full, in which case there probably isn't much you can do.

client/conn.go
```diff
@@ -0,0 +1,225 @@
...
+
+	addr          string
+	connTimeout   time.Duration
+	writeTimeout  time.Duration
+	keepAlive     bool
+	initThreshold int
+	multiplier    int
+	maxThreshold  int
+	maxDuration   time.Duration
+	nowFn         clock.NowFn
```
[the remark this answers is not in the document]
xichen2020: Yeah I can move it down.

client/conn.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	c.conn = tcpConn
+	return nil
+}
+
+func (c *connection) checkReconnectWithLock() error {
+	// If we haven't accumulated enough failures to warrant another reconnect
+	// and we haven't past the maximum duration since the last time we attempted
+	// to connect then we simply return false without reconnecting.
+	tooManyFailures := c.numFailures > c.threshold
```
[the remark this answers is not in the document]
xichen2020: Yeah see #34 , the main purpose here is to try to recover from a failed server quickly.

client/queue.go
```diff
@@ -0,0 +1,252 @@
...
+
+	"github.com/m3db/m3cluster/placement"
+	"github.com/m3db/m3metrics/encoding/protobuf"
+	"github.com/m3db/m3x/log"
+
+	"github.com/uber-go/tally"
+)
+
+const (
+	defaultQueueSizeNumBuckets = 6
```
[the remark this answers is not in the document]
xichen2020: Sure I think @jeromefroe added this so I can check with him.

client/conn.go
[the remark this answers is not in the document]
cw9: what's a typical value for `threshold` and `maxDuration`

client/writer.go
cw9: seems `enqueueErr` is only showing up in this code path and `encodeErr` is shared among all the code paths, so it's probably cleaner to handle enqueueErr here and encodeErr outside? but I don't feel strongly either way.

client/writer.go
cw9: Sounds good, please make sure there are test cases to guard that neither lock operation could be removed.

client/writer.go
cw9: right, but the buffer that failed to be enqueued here will just be discarded I guess?

client/writer.go
cw9: right this is still in the loop, make sense

client/conn.go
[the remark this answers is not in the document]
xichen2020: The threshold is dynamically adjusted as the number of failures accumulate, the default settings are:
```
	defaultInitReconnectThreshold       = 2
	defaultMaxReconnectThreshold        = 5000
	defaultReconnectThresholdMultiplier = 2
```
Will need to check on the `maxDuration` config.

client/writer.go
xichen2020: Yeah this was tested in `TestWriterConcurrentWriteStress` in `writer_test.go`.

client/writer.go
xichen2020: Correct, if the queue is full, the buffer will be discarded during the enqueue operation.

PR #104, opened 2018-04-24
Remove lint from build target

xichen2020:
cc @cw9 @jeromefroe 

This PR removes `lint` from the build targets since we seem to be getting the error below when installing golint. As such we remove `lint` and rely on `metalint` for linting. 
```
package golang.org/x/lint: unrecognized import path "golang.org/x/lint" (parse https://golang.org/x/lint?go-get=1: no go-import meta tags ())
```


xichen2020:
Decided to enable golint in metalinter config, hey it works!

PR #105, opened 2018-05-01
Admin client API for writing forwarded metrics

xichen2020:
cc @cw9 @jeromefroe 

This PR adds the admin client interface as well as the API for writing forwarded metrics to destination aggregation servers.

There are other minor renaming changes due to consolidating metric types in `m3metrics` but most of them are immaterial.

### DIFF
--- a/aggregator/capture/aggregator.go
+++ b/aggregator/capture/aggregator.go
@@ -26,6 +26,8 @@ import (
 
 	aggr "github.com/m3db/m3aggregator/aggregator"
 	"github.com/m3db/m3metrics/metadata"
+	"github.com/m3db/m3metrics/metric"
+	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3metrics/metric/id"
 	"github.com/m3db/m3metrics/metric/unaggregated"
 	"github.com/m3db/m3metrics/policy"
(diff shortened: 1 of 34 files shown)


aggregator/capture/aggregator.go
```diff
@@ -87,6 +89,14 @@ func (agg *aggregator) AddUntimed(
 	return nil
 }
 
+// TODO(xichen): implement this.
+func (agg *aggregator) AddForwarded(
+	metric aggregated.Metric,
+	metadata metadata.ForwardMetadata,
+) error {
+	return nil
```
jeromefroe: nit: would it be better to return a not implemented error just to be safe?

aggregator/list.go
```diff
@@ -354,18 +354,18 @@ func (l *metricList) discardLocalMetric(
...
 }
 
-// consumeForwardMetric consumes a forward metric.
+// consumeForwardedMetric consumes a forward metric.
 // TODO(xichen): implement this.
-func (l *metricList) consumeForwardMetric(
+func (l *metricList) consumeForwardedMetric(
 	metric aggregated.Metric,
 	meta metadata.ForwardMetadata,
 ) {
```
jeromefroe: nit: would it be worth panic'ing here just to ensure these functions aren't called until they are implemented?

aggregator/capture/aggregator.go
xichen2020: Sure can do, I'll add the implementation in the next PR but doesn't hurt either.

aggregator/list.go
xichen2020: Sure can do, I'll add the implementation in the next PR but doesn't hurt either.

PR #106, opened 2018-05-14
Add aggregator logic to write forwarded metrics

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to write forwarded metrics.

I left a few TODOs in the PR to remind myself of places that require a bit more work to make it work with forwarded metrics. They are mainly related to leadership election logic, and I intend to address the TODOs in a followup PR, but this PR is ready for review as is.


aggregator/generic_elem.go
```diff
@@ -88,14 +95,24 @@ type typeSpecificElemBase interface {
 	Close()
 }
 
+type lockedAggregation struct {
```
cw9: rename to `lockableAggregation`?

aggregator/list.go
```diff
@@ -85,88 +128,81 @@ func newMetricListMetrics(scope tally.Scope) metricListMetrics {
...
 	if err != nil {
 		return nil, err
 	}
 
-	// NB(xichen): by default the flush interval is the same as metric
-	// resolution, unless the resolution is smaller than the minimum flush
-	// interval, in which case we use the min flush interval to avoid excessing
-	// CPU overhead due to flushing.
-	flushInterval := resolution
-	if minFlushInterval := opts.MinFlushInterval(); flushInterval < minFlushInterval {
```
cw9: you don't need this MinFlushInterval logic any more?

aggregator/list.go
```diff
@@ -335,66 +381,255 @@ func (l *metricList) consumeLocalMetric(
...
+	flushMgr FlushManager
+}
+
+func newStandardMetricList(
+	shard uint32,
+	id standardMetricListID,
+	opts Options,
+) (*standardMetricList, error) {
+	var (
+		isEarlierThanFn  = isStandardMetricEarlierThan
```
cw9: Directly use the 2 functions?

aggregator/entry.go
```diff
@@ -440,72 +544,192 @@ func (e *Entry) updateMetadatasWithLock(
...
+
+	e.RLock()
+	if e.closed {
+		e.RUnlock()
+		timeLock.RUnlock()
+		return errEntryClosed
+	}
+
+	// Reject datapoints that arrive too late.
+	if err := e.checkLateness(
```
cw9: Could metrics with different `numForwardedTimes` being aggregated together? In that case is it ok for them to have different allowed lateness due to different `numForwardedTimes` in their metadata?

aggregator/generic_elem.go
```diff
@@ -175,16 +195,41 @@ func (e *GenericElem) ResetSetData(
...
+// AddUnique adds a metric value from a given source at a given timestamp.
+// If previous values from the same source have already been added to the
+// same aggregation, the incoming value is discarded.
+func (e *GenericElem) AddUnique(timestamp time.Time, value float64, sourceID uint32) error {
+	alignedStart := timestamp.Truncate(e.sp.Resolution().Window).UnixNano()
+	lockedAgg, err := e.findOrCreate(alignedStart)
+	if err != nil {
+		return err
+	}
+	lockedAgg.Lock()
```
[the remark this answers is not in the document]
xichen2020: See above for the rationale.

I could also rename it to just `agg` but it'll be `agg.aggregation.blah` instead of `lockedAgg.aggregation.blah` when referencing some of the nested data fields. Plus `lockedAgg` makes it more explicit that this is protected by a lock. Wdyt?

aggregator/list.go
xichen2020: Yeah I gave this more thought and realized that when the resolution is high (1s), chances are that users do want to see those metrics get flushed more quickly, so I got rid of the restriction here. We should probably add capacity or use a dedicated agg tier for those super high res metrics instead of using a longer flush interval since that kinda defeats the purpose of having high res metrics.

Additionally, this makes it difficult to reason what the max tolerated lateness should be given the metric resolution and number of forwarded times, since the resolution is not always the same as the flush interval.

aggregator/list.go
xichen2020: Sure

aggregator/entry.go
```diff
@@ -440,72 +544,192 @@ func (e *Entry) updateMetadatasWithLock(
...
+
+	e.Lock()
+	if e.closed {
+		e.Unlock()
+		timeLock.RUnlock()
+		return errEntryClosed
+	}
+
+	// Update metatadata, and add metric.
+	if e.shouldUpdateForwardMetadataWithLock(metadata) {
```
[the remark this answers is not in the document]
cw9: ah nvm I misread something

aggregator/entry.go
```diff
@@ -440,72 +544,192 @@ func (e *Entry) updateMetadatasWithLock(
...
+	}
+
+	err := e.addForwardedWithLock(metric, metadata.SourceID)
+	e.Unlock()
+	timeLock.RUnlock()
+
+	return err
+}
+
+func (e *Entry) checkLateness(
```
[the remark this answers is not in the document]
cw9: I see, I don't feel strongly either way in that case

aggregator/entry.go
cw9: sg

aggregator/generic_elem.go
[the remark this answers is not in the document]
cw9: Sure, I'm fine either way

aggregator/generic_elem.go
cw9: sg

PR #109, opened 2018-05-21
Use metric id as the source id and add integration tests

xichen2020:
cc @cw9 @jeromefroe 

This PR replaces the source server ID with the source metric ID as the source ID for forwarded metrics. This is needed because a single server can have multiple raw metrics matching the same rollup rule and thus each of them produces the same forwarded metric ID. As such, we can't use the shard set id as the source id for forwarded metrics (otherwise only one metric will be accepted and the rest will be rejected). As a result, this PR changes it to use the source metric ID that produces the forward metric ID as the source ID as the unique identifier during leader re-elections.

This PR also adds integration tests for better e2e coverage.


aggregator/generic_elem.go
```diff
@@ -277,9 +279,23 @@ func (e *GenericElem) Consume(
...
+		e.toConsume[i].lockedAgg.Lock()
+		e.processValueWithAggregationLock(timeNanos, e.toConsume[i].lockedAgg, flushLocalFn, flushForwardedFn)
 		// Closes the aggregation object after it's processed.
+		e.toConsume[i].lockedAgg.closed = true
 		e.toConsume[i].lockedAgg.aggregation.Close()
+		if e.toConsume[i].lockedAgg.sourcesSeen != nil {
+			sourceSetSize := len(e.toConsume[i].lockedAgg.sourcesSeen)
+			if sourceSetSize <= maxCachedSourceSetSize {
+				e.cachedSourceSetsLock.Lock()
+				if len(e.cachedSourceSets) < maxNumCachedSourceSets {
```
cw9: what happens if len(e.cachedSourceSets) >= maxNumCachedSourceSets? why not keep appending?

xichen2020: This is to make sure we don't take up too much space caching the source sets. Usually there will be two live source sets: one that's just finished and ready to be consumed, and the other that's being actively added to. Can tune more if needed.

cw9: kk, maybe just add some comments for this? Also why is this logic only executed when `sourceSetSize <= maxCachedSourceSetSize`?

xichen2020: Sure yeah can add comments.

In a similar vein, if the source set we are about to cache is too big (i.e., bigger than the configured max size), we do not want to retain it and instead simply discard it (this is useful if say over the course of a month the raw metric IDs producing the forward metric ID has changed significantly so there are a lot of stale ids in the map so we regenerate the map once in a while to refresh the IDs in the map).

PR #110, opened 2018-06-06
Pull in latest m3metrics dev

xichen2020:
cc @cw9 

This PR pulls in the latest m3metrics on tip of dev. It also initializes two bytes pool that weren't properly initialized previously.

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -190,8 +190,9 @@ func (c *EncoderConfiguration) NewEncoderOptions(
 	if c.BytesPool != nil {
 		sizeBuckets := c.BytesPool.NewBuckets()
 		objectPoolOpts := c.BytesPool.NewObjectPoolOptions(instrumentOpts)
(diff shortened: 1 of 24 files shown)


client/config.go
```diff
@@ -190,8 +190,9 @@ func (c *EncoderConfiguration) NewEncoderOptions(
 	if c.BytesPool != nil {
 		sizeBuckets := c.BytesPool.NewBuckets()
 		objectPoolOpts := c.BytesPool.NewObjectPoolOptions(instrumentOpts)
-		bytesPoolOpts := pool.NewBytesPool(sizeBuckets, objectPoolOpts)
-		opts = opts.SetBytesPool(bytesPoolOpts)
+		bytesPool := pool.NewBytesPool(sizeBuckets, objectPoolOpts)
+		opts = opts.SetBytesPool(bytesPool)
+		bytesPool.Init()
```
cw9: hmm, the bytes pool was never initialized?

xichen2020: Nope, so it panic'ed on startup.

PR #111, opened 2018-06-06
Add client metrics for better visibility

xichen2020:
cc @cw9 @jeromefroe 

This PR adds a more comprehensive set of client-side metrics to allow us to gain more visibility.

### DIFF
--- a/client/queue.go
+++ b/client/queue.go
@@ -118,15 +118,20 @@ type queue struct {
 	bufCh    chan protobuf.Buffer
 	doneCh   chan struct{}
 	closed   bool
+	wg       sync.WaitGroup
 
 	writeFn writeFn
 }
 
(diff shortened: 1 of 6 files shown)


client/queue.go
```diff
@@ -139,8 +144,10 @@ func newInstanceQueue(instance placement.Instance, opts Options) instanceQueue {
 	}
 	q.writeFn = q.conn.Write
 
+	q.wg.Add(2)
```
jeromefroe: What do you think of using an anonymous function to make it clear when `wg.Done` is closed? For example:

```
go func() {
  q.drain()
  q.wg.Done()
}
```

xichen2020: Talked offline, will keep it as is to be consistent with other parts of the code base.

PR #123, opened 2018-06-08
Add m3msg based router

cw9:
So users can config a router based on m3msg producer instead of using statically configured list of servers.


aggregator/handler/config.go
```diff
@@ -139,10 +155,48 @@ type flushHandlerConfiguration struct {
 	Type Type `yaml:"type"`
 
 	// Backend configures the backend.
-	Backend *backendConfiguration `yaml:"backend"`
+	Backend *staticBackendConfiguration `yaml:"backend"`
```
xichen2020: If you structure the config this way, I think `Type` above should be part of `staticBackendConfiguration` instead since that only applies to static backends. IOW,
```
type flushHandlerConfiguration struct {
  StaticBackend *staticBackendConfiguration `yaml:"staticBackend"`
  DynamicBackend *dynamicBackendConfiguration `yaml:"dynamicBackend"`
}
```
And which one is valid depends on which pointer is configured.

aggregator/handler/config.go
```diff
@@ -140,6 +155,47 @@ type flushHandlerConfiguration struct {
...
+
+type m3msgBackendConfiguration struct {
+	// Name of the backend.
+	Name string `yaml:"name"`
+
+	// Hashing function type.
+	HashType sharding.HashType `yaml:"hashType"`
+
+	// Total number of shards.
+	TotalShards int `yaml:"totalShards" validate:"nonzero"`
```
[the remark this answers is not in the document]
xichen2020: I see sounds good.

aggregator/handler/router/allow_all.go
```diff
@@ -0,0 +1,18 @@
+package router
```
xichen2020: License.

aggregator/handler/router/broadcast.go
```diff
@@ -0,0 +1,33 @@
+package router
```
xichen2020: License.

aggregator/handler/router/router.go
```diff
@@ -0,0 +1,13 @@
+package router
```
xichen2020: License.

aggregator/handler/router/sharded.go
```diff
@@ -0,0 +1,68 @@
+package router
```
xichen2020: License.

aggregator/handler/router/with_ack.go
```diff
@@ -0,0 +1,48 @@
...
+func (r withAckRouter) Close() {
+	r.p.Close(producer.WaitForConsumption)
+}
+
+type message struct {
+	shard  uint32
+	buffer *common.RefCountedBuffer
+}
+
+func newMessage(shard uint32, buffer *common.RefCountedBuffer) producer.Message {
```
xichen2020: This leads to one allocation per call due to casting to the interface, do you need to pool this?

aggregator/handler/router/with_ack.go
```diff
@@ -0,0 +1,48 @@
+package router
```
xichen2020: License.

aggregator/handler/router/with_ack_test.go
```diff
@@ -0,0 +1,22 @@
+package router
```
xichen2020: License.

aggregator/handler/type.go
```diff
@@ -33,13 +33,15 @@ const (
 	blackholeType Type = "blackhole"
 	loggingType   Type = "logging"
 	forwardType   Type = "forward"
+	dynamicType   Type = "dynamic"
```
xichen2020: As mentioned above, you don't need this type if you change how the flush handler config is structured.

aggregator/options.go
```diff
@@ -62,6 +62,7 @@ var (
 	// are issues with the instances taking over the shards and as such we need to switch
 	// the traffic back to the previous owner of the shards immediately.
 	defaultBufferDurationAfterShardCutoff = time.Hour
+	defaultCloseTimeout                   = 5 * time.Second
```
xichen2020: As mentioned above, let's use the `gracefulShutdownTimeout` for this purpose instead of having a second shutdown timeout.

services/m3aggregator/config/aggregator.go
```diff
@@ -138,6 +138,9 @@ type AggregatorConfiguration struct {
 
 	// Pool of entries.
 	EntryPool pool.ObjectPoolConfiguration `yaml:"entryPool"`
+
+	// CloseTimeout configurates the timeout for closing the aggregator.
+	CloseTimeout *time.Duration `yaml:"closeTimeout"`
```
xichen2020: As mentioned above, no need for this.

aggregator/aggregator.go
```diff
@@ -214,6 +215,25 @@ func (agg *aggregator) Status() RuntimeStatus {
 }
 
 func (agg *aggregator) Close() error {
+	var (
```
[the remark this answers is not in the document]
cw9: Sure, I wasn't aware of this graceful shutdown, that should be sufficient

aggregator/handler/config.go
cw9: sg

aggregator/handler/router/with_ack.go
cw9: The message pool in producer creates 3X(3 consumers) objects whenever one of this message here was created, and the result shows using allocation there is more efficient than pooling, so I feel we probably don't need to pool this.

aggregator/handler/config.go
```diff
@@ -69,34 +69,36 @@ func (c FlushHandlerConfiguration) NewHandler(
...
+				cs,
+				instrumentOpts,
+			)
+			if err != nil {
+				return nil, err
+			}
+			sharderRouters = append(sharderRouters, sharderRouter)
+			continue
+		}
+		switch hc.StaticBackend.Type {
```
xichen2020: Can we move this `switch` statement to the `NewSharderRouter` method of `StaticBackend` so it's a bit cleaner?

aggregator/handler/router/with_ack.go
xichen2020: I think pooling is still useful when the number of objects needed / s is not too high and there aren't a large number of goroutines trying to get them at the same time. Empirically speaking the agg tier produces roughly 8M metrics / s, so roughly 800K buffers / s, given the agg tier size, that's roughly 4K buffers / s per instance, IOW you'll have about 16 goroutines contending for a total of 4K buffers / s during flushing, which is not a lot of contention which is why I think pooling might actually help here.

You might be right that that's still too much contention overhead for the agg tier, but I'd like to see some empirical evidence with production agg tier load as opposed to microbenchmarks. If you intend to measure this in the future, can you either create an issue and/or add a TODO here?

aggregator/handler/config.go
cw9: I could, but it may create a handler or a SharderRouter depends on the type, so I'll need a `NewHandlerOrSharderRouter` function and return 2 arguments, then do the append here, which is also messy, up to you.

aggregator/handler/router/with_ack.go
cw9: Sure, can add a pool here if needed, will add a TODO for now.

aggregator/handler/config.go
xichen2020: Hm we could pass in the handlers and routers, do the append inside the method, and return them, but yeah don't feel strongly about it, let's leave it as is for now.

aggregator/handler/router/with_ack.go
```diff
@@ -47,6 +47,7 @@ type message struct {
 	buffer *common.RefCountedBuffer
 }
 
+// TODO: Pool the messages if needed.
```
xichen2020: nit: can you assign this to you (i.e., making it `TODO (cw9)`)? I usually prefer to put a name next to the TODOs or otherwise the ownership of TODOs is not really clear when I clean them up later.

xichen2020:
Can you also rerun the tests to make sure they pass? Seems glide / git flakiness.

aggregator/handler/config.go
```diff
@@ -135,14 +153,62 @@ func (c *writerConfiguration) NewWriterOptions(
...
+
+	// DynamicBackend configures the dynamic backend.
+	DynamicBackend *dynamicBackendConfiguration `yaml:"dynamicBackend"`
+}
 
-	// Backend configures the backend.
-	Backend *backendConfiguration `yaml:"backend"`
+func (c flushHandlerConfiguration) Validate() error {
+	if c.StaticBackend == nil && c.DynamicBackend == nil {
+		return errors.New("neither dynamic nor static backend was configured")
```
xichen2020: Can we make this a global var?

aggregator/handler/config.go
```diff
@@ -135,14 +153,62 @@ func (c *writerConfiguration) NewWriterOptions(
...
+	}
+	if c.StaticBackend == nil && c.DynamicBackend == nil {
```
xichen2020: This seems like it should be `if c.StaticBackend != nil && c.DynamicBackend != nil {`, can we add a test?

aggregator/handler/config.go
```diff
@@ -135,14 +153,62 @@ func (c *writerConfiguration) NewWriterOptions(
...
+		return errors.New("both static and dynamic backend were configured")
```
xichen2020: Can we make this a global var?

aggregator/handler/router/with_ack.go
```diff
@@ -0,0 +1,69 @@
...
+func (r withAckRouter) Close() {
+	r.p.Close(producer.WaitForConsumption)
+}
+
+type message struct {
+	shard  uint32
+	buffer *common.RefCountedBuffer
+}
+
+// TODO: Pool the messages if needed.
```
xichen2020: Add an owner for the TODO.

aggregator/handler/config.go
cw9: good catch, will do

aggregator/handler/config.go
```diff
@@ -45,9 +45,10 @@ const (
 )
 
 var (
-	errNoHandlerConfiguration = errors.New("no handler configuration")
-	errNoWriterConfiguration  = errors.New("no writer configuration")
-	errNoBackendConfiguration = errors.New("no backend configuration")
+	errNoHandlerConfiguration                   = errors.New("no handler configuration")
+	errNoWriterConfiguration                    = errors.New("no writer configuration")
+	errNoDynamicNorStaticBackendConfiguration   = errors.New("neither dynamic nor static backend was configured")
```
xichen2020: nit: `errNoDynamicOrStaticBackendConfiguration`

aggregator/handler/config.go
```diff
@@ -197,12 +202,26 @@ func (c *dynamicBackendConfiguration) NewSharderRouter(
 	if err := p.Init(); err != nil {
 		return SharderRouter{}, err
 	}
+	for _, filter := range c.Filters {
+		sid, f := filter.NewConsumerServiceFilter()
+		p.RegisterFilter(sid, f)
+		instrumentOpts.Logger().Infof("registered filter for consumer service: %s", sid.String())
```
xichen2020: nit: move the logger outside the for loop.

Also do you need this log message? is it too verbose?

Also do you want to implement `String()` for the filter?

aggregator/handler/router/with_ack_test.go
```diff
@@ -23,6 +23,8 @@ package router
 import (
 	"testing"
 
+	"github.com/m3db/m3aggregator/sharding"
+
```
xichen2020: Remove the blank line.

aggregator/handler/router/with_ack_test.go
```diff
@@ -31,12 +33,22 @@ import (
...
 	require.Equal(t, uint32(2), msg.Shard())
-	require.Equal(t, uint32(0), msg.Size())
+	require.Equal(t, 1024, msg.Size())
 	require.Empty(t, msg.Bytes())
 
 	msg.Finalize(producer.Consumed)
 	require.Panics(t, buf.DecRef)
 }
+
+func TestM3msgShardSetFilter(t *testing.T) {
```
xichen2020: Rename `TestM3msg` to `TestWithAckRouter`

aggregator/handler/router/with_ack.go
```diff
@@ -60,10 +61,27 @@ func (d message) Bytes() []byte {
 	return d.buffer.Buffer().Bytes()
 }
 
-func (d message) Size() uint32 {
-	return uint32(len(d.Bytes()))
+func (d message) Size() int {
+	// Use the cap of the underlying byte slice in the buffer instead of
+	// the length of the byte encoded.
+	return d.buffer.Buffer().Buffer().Cap()
```
xichen2020: nit: `cap(d.Bytes())`?

aggregator/handler/router/with_ack.go
```diff
@@ -60,10 +61,27 @@ func (d message) Bytes() []byte {
...
 }
 
 func (d message) Finalize(producer.FinalizeReason) {
 	d.buffer.DecRef()
 }
+
+type shardFilter struct {
```
xichen2020: Can you move this and the ctor etc to a `shard.go` under `aggregator/handler/filter` package?

aggregator/handler/router/with_ack.go
```diff
@@ -60,10 +61,27 @@ func (d message) Bytes() []byte {
 	return d.buffer.Buffer().Bytes()
 }
 
-func (d message) Size() uint32 {
-	return uint32(len(d.Bytes()))
+func (d message) Size() int {
+	// Use the cap of the underlying byte slice in the buffer instead of
+	// the length of the byte encoded.
```
xichen2020: Can you also explain a bit more on why the cap instead of the length is the right value to use?

aggregator/handler/config.go
cw9: It's probably fine since it only print once per each defined filter, usually we will only have one of this, I feel it's nice to have something to indicate the filter is being applied.

PR #135, opened 2018-06-29
Allow multiple aggregation keys for forwarded metrics

xichen2020:
cc @jeromefroe @cw9 

This PR adds logic to allow multiple aggregation keys for forwarded metrics.

### DIFF
--- a/aggregator/entry.go
+++ b/aggregator/entry.go
@@ -616,8 +616,14 @@ func (e *Entry) addForwarded(
 	}
 
 	// Check if we should update metadata, and add metric if not.
-	if !e.shouldUpdateForwardMetadataWithLock(metadata) {
-		err := e.addForwardedWithLock(metric, metadata.SourceID)
+	key := aggregationKey{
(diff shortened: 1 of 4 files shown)


aggregator/entry.go
```diff
@@ -612,8 +616,14 @@ func (e *Entry) addForwarded(
 	}
 
 	// Check if we should update metadata, and add metric if not.
-	if !e.shouldUpdateForwardMetadataWithLock(metadata) {
-		err := e.addForwardedWithLock(metric, metadata.SourceID)
+	key := aggregationKey{
```
cw9: Maybe make a constructor for new aggregationKey and reuse in all the places? in case you added a field to the struct and forgot to assign value in some places.

xichen2020: Talked offline, decided to keep this as is due to its limited use in `entry.go` and `elem_base.go` and convenience to not have to provide all values and can rely on default values for initialization.

PR #139, opened 2018-07-05
Add traffic control to the flush backends

cw9:
This pr adds a traffic controller that can be configured per flush backend to enable/disable traffic.

@xichen2020


aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
xichen2020: 2018

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
...
+// TrafficControlOptions configurates the traffic control options.
+type TrafficControlOptions interface {
+	// SetStore sets the kv store.
+	SetStore(store kv.Store) TrafficControlOptions
+
+	// Store returns the kv store.
+	Store() kv.Store
+
+	// SetDefaultDisabled sets the default disabled value.
+	SetDefaultDisabled(value bool) TrafficControlOptions
```
xichen2020: Would it be more preferable to use `SetDefaultEnabled` instead? I assume you want to rely on the default value of `false` to enable traffic by default, but it's a bit mind-twisting to think about what it means when setting disabled to false.

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// DefaultDisabled returns the default disabled value.
+	DefaultDisabled() bool
+
+	// SetRuntimeDisableKey sets the runtime disable key.
+	SetRuntimeDisableKey(value string) TrafficControlOptions
```
xichen2020: Same here

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	o.instrumentOpts = value
+	return &o
+}
+
+func (opts *trafficControlOptions) InstrumentOptions() instrument.Options {
+	return opts.instrumentOpts
+}
+
+// TrafficController controls if traffic is enabled.
+type TrafficController struct {
```
xichen2020: Have you considered using the runtime manager in `runtime` package for controlling the runtime behavior? This is an established pattern in this repo where the objects wishing to listen to runtime config changes register with the runtime manager and take actions on updates, and the runtime manager manages watching the keys and updating the watchable to send notifications to the watchers, which is a very clean divide of responsibility.

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	disabled *atomic.Bool
+}
+
+// NewTrafficController creates a new traffic controller.
+func NewTrafficController(opts TrafficControlOptions) *TrafficController {
+	disabled := atomic.NewBool(opts.DefaultDisabled())
```
xichen2020: Should this belong to an `Start` or `Init` method on the traffic controller?

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil
+	}
+	vOptions := watch.NewOptions().
+		SetInitWatchTimeout(opts.InitTimeout()).
+		SetInstrumentOptions(iOpts).
+		SetNewUpdatableFn(newUpdatableFn).
+		SetGetUpdateFn(getFn).
+		SetProcessFn(processFn)
+	value := watch.NewValue(vOptions)
+	value.Watch()
```
xichen2020: Probably should have a `Close` method on the traffic controller as well to unwatch the runtime value or otherwise this leaks goroutines.

aggregator/handler/config.go
```diff
@@ -171,6 +175,41 @@ func (c flushHandlerConfiguration) Validate() error {
 	return nil
 }
 
+type trafficControlConfiguration struct {
+	DefaultDisabled   bool                     `yaml:"defaultDisabled" validate:"nonzero"`
```
xichen2020: Hm validate nonzero here means this can't be set to false in the config, that's probably not what you want? It's probably better to use a `*bool` and check for nil explicitly.

aggregator/handler/config.go
```diff
@@ -171,6 +175,41 @@ func (c flushHandlerConfiguration) Validate() error {
...
+	RuntimeDisableKey string                   `yaml:"runtimeDisableKey" validate:"nonzero"`
+	InitTimeout       *time.Duration           `yaml:"initTimeout"`
+	KVOverride        kv.OverrideConfiguration `yaml:"kvOverride"`
```
xichen2020: Do you really need the kv override per traffic control config? I'd imagine these would be the same for all the backends you want to control traffic for. If so, can probably just pass in the store to the ctor and get rid of this.

aggregator/handler/config.go
```diff
@@ -171,6 +175,41 @@ func (c flushHandlerConfiguration) Validate() error {
 	return nil
 }
 
+type trafficControlConfiguration struct {
```
xichen2020: nit: move this to be after `staticBackendConfiguration` struct definition to roughly match the ordering in which the types appear.

aggregator/handler/router/traffic_controlled.go
```diff
@@ -0,0 +1,52 @@
...
+) Router {
+	return &trafficControlledRouter{
+		TrafficController: trafficController,
+		Router:            router,
+	}
+}
+
+func (r *trafficControlledRouter) Route(shard uint32, buffer *common.RefCountedBuffer) error {
+	if !r.TrafficController.Allow() {
+		return nil
```
xichen2020: You need to decrement the buffer ref here or otherwise the buffers won't be returned to pool.

Also not sure if you want to add some metrics here.

aggregator/handler/router/traffic_controlled.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	return r.Router.Route(shard, buffer)
+}
+
+func (r *trafficControlledRouter) Close() {
+	r.Router.Close()
```
xichen2020: This is a good opportunity to close the traffic controller if it has a `Close` method.

aggregator/handler/router/traffic_controlled_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+package router
+
+import (
+	"testing"
+	"time"
+
+	"github.com/m3db/m3cluster/kv/mem"
+
+	"github.com/golang/mock/gomock"
```
xichen2020: Import ordering.

aggregator/handler/router/traffic_controlled.go
cw9: good catch, will do

aggregator/handler/common/traffic_controller.go
xichen2020: Agreed it's more involved than watching a single key but I think it's not too difficult if you give a unique ID to each backend (e.g., "ingester-short-static" for the legacy one and "ingester-short-dynamic" for the new one). Then you can set one key per backend and distinguish the keys among the backends similarly to how the runtime manager keeps track of keys.

I guess it's fine to stick w/ the kv based approach if this is going to live on a branch.

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,205 @@
...
+// Init initializes the traffic controller to watch the runtime updates.
+func (c *TrafficController) Init() error {
+	err := c.value.Watch()
+	if err == nil {
+		return nil
+	}
+	if _, ok := err.(watch.CreateWatchError); ok {
+		return err
+	}
+	return nil
```
xichen2020: What other errors can we return here?

cw9: You mean from the value.Watch()? It can also be `InitValueError`.

aggregator/handler/common/traffic_controller.go
cw9: Sure, I thought about that, but that makes the runtime value update logic rather complex/unituitive, we'd have to define runtime configs for some of the backends at the aggregator level, far away from the backend configuration itself which is not great. 
I guess the main difference here is that the current keys in the runtime options are expected for the aggregator, but the traffic control logic is optional per backend.

aggregator/handler/common/traffic_controller.go
xichen2020: I see, do you want to be safe here and just return the error? This is usually because something is wrong with KV, in which case relying on the default value may not be safe, so I'm not sure if you want to proceed here when the error occurs.

aggregator/handler/common/traffic_controller.go
xichen2020: Yeah I've been meaning to move the handler outside the aggregator package since the interfaces are cleanly defined and separated at the moment. After that, you could leave the existing runtime manager for the aggregator as is (maybe move it into the aggregator package), and create a new runtime manager just for the handlers (maybe keep it in the handler package), in which case it'll make more sense to configure the backend keys for the handler runtime manager.

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,280 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
cw9: sg

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	// Close closes the traffic controller.
+	Close()
+}
+
+// TrafficControllerType defines the type of the traffic controller.
+type TrafficControllerType string
+
+// Supported types.
+var (
+	Enabler  TrafficControllerType = "enabler"
```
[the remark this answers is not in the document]
cw9: sg

aggregator/handler/common/traffic_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	// SetInstrumentOptions sets the instrument options.
+	SetInstrumentOptions(value instrument.Options) TrafficControlOptions
+
+	// InstrumentOptions returns the instrument options.
+	InstrumentOptions() instrument.Options
+}
+
+type trafficControlOptions struct {
+	store          kv.Store
+	defaultValue   bool
```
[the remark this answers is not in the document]
cw9: adding the defaultvalue back to the options

aggregator/handler/router/traffic_control_options.go
```diff
@@ -84,9 +64,10 @@ func (t *TrafficControllerType) UnmarshalYAML(unmarshal func(interface{}) error)
 
 // TrafficControllerConfiguration configures the traffic controller.
 type TrafficControllerConfiguration struct {
-	Type        TrafficControllerType `yaml:"type"`
-	RuntimeKey  string                `yaml:"runtimeKey" validate:"nonzero"`
-	InitTimeout *time.Duration        `yaml:"initTimeout"`
+	Type         TrafficControllerType `yaml:"type"`
+	DefaultValue bool                  `yaml:"defaultValue"`
```
xichen2020: Let's make this a `*bool` so you don't accidentally set this to false in the options when you forget to set it in the config.

aggregator/handler/router/traffic_control_options.go
```diff
@@ -18,48 +18,28 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package common
+package router
```
xichen2020: Looking at the changes, it appears it would simplify things a lot if all traffic control related files are in its own package under `handler/router/trafficcontrol/`, then you can have `options.go`, `config.go`, `router.go`, `controller.go` etc and remove `TrafficControl` from the names of traffic-control related types and methods.

Thoughts?

aggregator/handler/router/traffic_control_options.go
```diff
@@ -153,128 +141,57 @@ type trafficControlOptions struct {
 // NewTrafficControlOptions creats a new TrafficControlOptions.
 func NewTrafficControlOptions() TrafficControlOptions {
 	return &trafficControlOptions{
-		defaultValue:   defaultDefaultDisabled,
```
xichen2020: No initial value for `defaultValue`?

aggregator/handler/router/traffic_control_options.go
cw9: sure, sg

aggregator/handler/router/traffic_control_options.go
cw9: It will just be false, can set explicitly

aggregator/handler/router/traffic_control_options.go
cw9: sg

PR #140, opened 2018-07-06
Do not flush NaN values downstream optionally

xichen2020:
cc @jeromefroe @cw9 

This PR provides the logic to optionally discard NaN aggregated values and do not forward them to other servers for further aggregation or flush them to downstream services for indexing and storage.

Forwarding NaN values to other aggregation servers as part of the rollup pipeline will lead to NaN rollup metric values, which are currently treated as no datapoints in our query engine and therefore cause the entire rollup metric value to go "missing". Instead, we can optionally drop NaN values at the source so the multi-level rollup metrics are not affected as long as the remaining constituent metric values are non-NaN values.

### DIFF
--- a/aggregator/counter_elem.gen.go
+++ b/aggregator/counter_elem.gen.go
@@ -27,6 +27,8 @@ package aggregator
 import (
 	"fmt"
 
+	"math"
+
 	"sync"
 
 	"time"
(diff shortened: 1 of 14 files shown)


aggregator/counter_elem.gen.go
```diff
@@ -27,6 +27,8 @@ package aggregator
 import (
 	"fmt"
 
+	"math"
+
```
jeromefroe: newlines?

aggregator/counter_elem.gen.go
```diff
@@ -417,7 +420,9 @@ func (e *CounterElem) processValueWithAggregationLock(
 				value = res.Value
 			}
 		}
-		// Do we need to flush or forward NaNs?
+		if discardNaNValues && math.IsNaN(value) {
```
jeromefroe: nit: should this condition be the other way? ie we only need to check `discardNaNValues` if `value` is a NaN

aggregator/counter_elem.gen.go
xichen2020: Yeah unfortunately these are auto-generated by genny. I can take a look to see if it's easy to clean up the imports when I have some down time.

aggregator/counter_elem.gen.go
xichen2020: Yeah I imagine checking `discardNaNValues` is probably cheaper than checking if a value is NaN or not, and if so can short circuit early if the `discardNaNValues` is false.

PR #141, opened 2018-07-08
Reset next flush time when new flusher is registered

xichen2020:
cc @cw9 @jeromefroe 

This PR adds logic to reset bucket next flush time to establish a limit on when a new flushing list will be flushed next.  This allows the follower to properly discard metrics and helps to speed up deployment in certain cases (e.g., when a leader just performed a flush and a new list is created immediately afterwards due to a new metric coming in with a new shard/resolution combination, and the follower wouldn't resign until the leader performs the next flush for the new list), both of which rely on the last flushed times of the flushers.

### DIFF
--- a/aggregator/leader_flush_mgr.go
+++ b/aggregator/leader_flush_mgr.go
@@ -192,6 +192,32 @@ func (mgr *leaderFlushManager) OnBucketAdded(
 	mgr.Unlock()
 }
 
+// Reset the next flush timestamp of the bucket to be t0 = truncate(now, flushInterval)
+// + flushOffset. As such, the newly added flusher will be flushed at t0 (or immediately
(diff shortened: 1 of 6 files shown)


aggregator/leader_flush_mgr.go
```diff
@@ -460,9 +495,10 @@ func (h flushMetadataHeap) heapify(i, n int) {
 			smallest = right
 		}
 		if smallest == i {
-			return
+			break
 		}
 		h[i], h[smallest] = h[smallest], h[i]
 		i = smallest
 	}
+	return i > i0
```
cw9: Add a comment on what's the returned value?

xichen2020: Sure updated.

PR #143, opened 2018-07-12
Eager flushing of forwarded metrics

xichen2020:
cc @cw9 @jeromefroe 

This PR adds support to enable eager flushing of forwarded metrics. With eager flushing enabled, the destination server where forwarded metrics are stored will build up the set of forwarding sources, and once the source set is built, it will use the source set to determine when to eagerly flush forwarded metrics once it has received data from all the expected sources therefore reducing full forwarding delay.

### DIFF
--- a/aggregator/forwarded_writer.go
+++ b/aggregator/forwarded_writer.go
@@ -30,6 +30,7 @@ import (
 	"github.com/m3db/m3metrics/metric"
 	"github.com/m3db/m3metrics/metric/aggregated"
 	"github.com/m3db/m3metrics/metric/id"
+	"github.com/m3db/m3x/clock"
 	xerrors "github.com/m3db/m3x/errors"
 
 	"github.com/uber-go/tally"
@@ -72,8 +73,8 @@ type forwardedMetricWriter interface {
 		aggKey aggregationKey,
 	) error
 
-	// Prepare prepares the writer for a new write cycle.
-	Prepare()
+	// Prepare prepares the writer for a new write cycle flushing for given timestamp.
+	Prepare(targetNanos int64)
 
 	// Flush flushes any data buffered in the writer.
 	Flush() error
@@ -131,8 +132,12 @@ func newForwardedWriterMetrics(scope tally.Scope) forwardedWriterMetrics {
 // associated with the (shard, listID) combination, which is used for value
 // deduplication during leadership re-elections on the destination server.
 type forwardedWriter struct {
-	shard  uint32
-	client client.AdminClient
+	shard                  uint32
+	client                 client.AdminClient
+	eagerForwardingEnabled bool
+	maxForwardingWindows   int
+	isEarlierThanFn        isEarlierThanFn
+	nowFn                  clock.NowFn
 
 	closed             bool
 	aggregations       map[idKey]*forwardedAggregation // Aggregations for each forward metric id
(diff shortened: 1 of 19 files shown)


aggregator/metric_type.go
```diff
@@ -0,0 +1,73 @@
...
+type IncomingMetricType int
+
+const (
+	// UnknownIncomingMetric is an incoming metric with unknown type.
+	UnknownIncomingMetric IncomingMetricType = iota
+
+	// StandardIncomingMetric is a standard (currently untimed) incoming metric.
+	StandardIncomingMetric
+
+	// ForwardedIncomingMetric is a forwarded incoming metric.
```
cw9: Do you think we will ever have untimed forwarded metrics?

services/m3aggregator/config/aggregator.go
```diff
@@ -276,6 +275,37 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	maxAllowedForwardingDelayFn := c.Forwarding.MaxAllowedForwardingDelayFn(jitterEnabled, maxJitterFn)
 	opts = opts.SetMaxAllowedForwardingDelayFn(maxAllowedForwardingDelayFn)
 
+	// Set whether to enable eager forwarding.
+	if c.Forwarding.EnableEagerForwarding != nil {
```
cw9: move these logic to a method on `forwardingConfiguration` struct?

services/m3aggregator/config/aggregator.go
```diff
@@ -436,7 +461,11 @@ func (c placementManagerConfiguration) NewPlacementManager(
 
 type forwardingConfiguration struct {
 	// MaxSingleDelay is the maximum delay for a single forward step.
-	MaxSingleDelay time.Duration `yaml:"maxSingleDelay"`
+	MaxSingleDelay         time.Duration                      `yaml:"maxSingleDelay"`
```
cw9: if this is required, add a yaml validation?

aggregator/forwarded_writer.go
```diff
@@ -456,12 +515,70 @@ func (agg *forwardedAggregation) onDone(key aggregationKey) error {
...
+		// in between so the destination server knows when it has received from all
+		// sources and can perform its flush as soon as possible.
+		var (
+			currWriteNanos = agg.byKey[idx].lastWriteNanos + resolutionNanos
+			bucketIdx      = 0
+		)
+		lastWriteNanos = agg.byKey[idx].lastWriteNanos
+		for agg.isEarlierThanFn(currWriteNanos, agg.targetFlushNanos) || bucketIdx < len(agg.byKey[idx].bucketsByTimeAsc) {
+			var compareResult int
+			if bucketIdx >= len(agg.byKey[idx].bucketsByTimeAsc) {
```
jeromefroe: nit: would it be cleaner to use a switch statement here?

aggregator/generic_elem.go
```diff
@@ -224,11 +242,29 @@ func (e *GenericElem) AddUnique(timestamp time.Time, values []float64, sourceID
...
+		// clear the source. If the corresponding bit is not set, it implies either
+		// a duplicate delivery from that source, or a new source that was not in
+		// the initial cloned source set.
+		if !lockedAgg.sourcesSet.Test(source) {
+			lockedAgg.Unlock()
+			return errDuplicateOrNewForwardingSource
+		}
+		lockedAgg.sourcesSet.Clear(source)
+		if lockedAgg.sourcesSet.None() {
+			lockedAgg.consumeState = readyToConsume
```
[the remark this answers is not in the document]
xichen2020: Sure, can add comments. If source is ready for using, but the server has not heard from all sources yet, the consumer state is `notReadyToConsume`, in which case the server will continue to wait till either all sources have sent data or max delay has passed.

aggregator/generic_elem.go
```diff
@@ -469,12 +539,78 @@ func (e *GenericElem) processValueWithAggregationLock(
...
+		return false
+	}
+	if e.buildingSourcesAtNanos == 0 {
+		return false
+	}
+	// NB: Allow TTL for the source set to build up.
+	return e.nowFn().UnixNano() >= e.buildingSourcesAtNanos+e.sourcesTTLNanos
+}
+
+func (e *GenericElem) maybeRefreshSourcesSetWithLock() {
```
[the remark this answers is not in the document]
xichen2020: Sure can clarify, `try...` means it'll always try to refresh, which is not really what this function does, this function may possibly update sources set but not necessarily.

aggregator/metric_type.go
xichen2020: I don't think so.

services/m3aggregator/config/aggregator.go
xichen2020: Hm yeah I can try.

services/m3aggregator/config/aggregator.go
xichen2020: It's not required, if unspecfied, it uses the default value in the options, I'll make it a `*time.Duration`.

PR #145, opened 2018-07-13
Followers should not consume forwarded metrics eagerly

xichen2020:
cc @cw9 

This PR adds logic so that followers only discard metrics based on the leader's schedule and do not consume forwarded metrics eagerly.

### DIFF
--- a/aggregator/flush.go
+++ b/aggregator/flush.go
@@ -42,10 +42,10 @@ type flushingMetricList interface {
 	LastFlushedNanos() int64
 
 	// Flush performs a flush for a given request.
-	Flush(req flushRequest)
+	Flush(req flushRequest, eagerForwardingMode eagerForwardingMode)
 
 	// DiscardBefore discards all metrics before a given timestamp.
-	DiscardBefore(beforeNanos int64)
+	DiscardBefore(beforeNanos int64, eagerForwardingMode eagerForwardingMode)
 }
 
 // fixedOffsetFlushingMetricList is a flushing metric list that flushes at fixed offset
@@ -69,6 +69,13 @@ type flushRequest struct {
 	// If nonzero, data between [now - bufferAfterCutoff, now) are buffered.
 	BufferAfterCutoff time.Duration
 }
+
+type eagerForwardingMode int
+
+const (
+	allowEagerForwarding eagerForwardingMode = iota
+	disAllowEagerForwarding
+)
 
 type flushType int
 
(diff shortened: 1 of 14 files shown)


aggregator/flush.go
```diff
@@ -70,6 +70,13 @@ type flushRequest struct {
 	BufferAfterCutoff time.Duration
 }
 
+type eagerForwardingMode int
+
+const (
+	allowEagerForwarding eagerForwardingMode = iota
```
cw9: nit: `eagerForwardingEnabled` and `eagerForwardingDisabled`?

aggregator/follower_flush_mgr_test.go
```diff
@@ -410,8 +410,8 @@ func TestFollowerFlushTaskRun(t *testing.T) {
 		i := i
 		flusher := NewMockflushingMetricList(ctrl)
 		flusher.EXPECT().
-			DiscardBefore(gomock.Any()).
-			Do(func(beforeNanos int64) {
+			DiscardBefore(gomock.Any(), disAllowEagerForwarding).
+			Do(func(beforeNanos int64, _ eagerForwardingMode) {
```
cw9: can verify the forwardingMode passed in here?

aggregator/generic_elem.go
```diff
@@ -347,7 +348,7 @@ func (e *GenericElem) Consume(
 		// We only attempt to consume if the outgoing metrics type is local instead of forwarded.
```
cw9: Do you need to update the comments here?

aggregator/flush.go
xichen2020: Hm that conflicts with the options `EagerForwardingEnabled` and `EagerForwardingDisabled`. This is to convey the meaning that based on the leader / follower role, a flush may be allowed or disallowed to consume metrics even when eager forwarded is enabled.

aggregator/generic_elem.go
xichen2020: Sure can do.

aggregator/follower_flush_mgr_test.go
xichen2020: Sure

aggregator/follower_flush_mgr_test.go
```diff
@@ -404,15 +404,19 @@ func TestFollowerFlushTaskRun(t *testing.T) {
...
 	flushers := make([]flushingMetricList, 3)
 	for i := 0; i < 3; i++ {
 		i := i
 		flusher := NewMockflushingMetricList(ctrl)
 		flusher.EXPECT().
 			DiscardBefore(gomock.Any(), disAllowEagerForwarding).
-			Do(func(beforeNanos int64, _ eagerForwardingMode) {
+			Do(func(beforeNanos int64, eagerForwardingMode eagerForwardingMode) {
 				flushedBefore[i] = beforeNanos
+				eagerForwardingModes[i] = eagerForwardingMode
```
cw9: I think you can just do `require` here as well

PR #153, opened 2018-08-28
Validate leader value in election

cw9:


aggregator/election_mgr.go
```diff
@@ -504,6 +508,20 @@ func (mgr *electionManager) verifyPendingFollower(watch watch.Watch) {
 				mgr.logError("leader has not changed", errLeaderNotChanged)
 				return errLeaderNotChanged
 			}
+			_, p, err := mgr.placementManager.Placement()
+			if err != nil {
+				mgr.metrics.verifyPlacementErrors.Inc(1)
+				mgr.logError("error getting placement", err)
+				return err
+			}
+			_, exist := p.Instance(leader)
```
jskelcy: Seems like this will just tell you if the host exists, but not if the `electionKey` for the which this host is the leader?

This would tell us if the leader is a valid host, but not if that host thinks it is the leader. Perhaps this might be helpful:

```
instance, exist := p.Instance(leader)
if !exists {
   // handle unknown host as leader
}
if instance.ShardSetID() != mgr.ShardSetID {
  // handle leader with different shardSetID
}

```

aggregator/election_mgr.go
```diff
@@ -504,6 +508,20 @@ func (mgr *electionManager) verifyPendingFollower(watch watch.Watch) {
...
+			if !exist {
+				mgr.metrics.verifyLeaderNotInPlacement.Inc(1)
+				err := fmt.Errorf("received invalid leader value: [%s], which is not available in placement", leader)
+				mgr.logger.Error(err.Error())
```
xichen2020: Use `mgr.logError` here to be consistent?

aggregator/election_mgr.go
```diff
@@ -504,6 +508,20 @@ func (mgr *electionManager) verifyPendingFollower(watch watch.Watch) {
...
+				return err
+			}
+			mgr.logger.Infof("found valid new leader: [%s] for the campaign", leader)
```
xichen2020: Do you need check if the current instance and the leader instance are in the same group?

aggregator/election_mgr.go
```diff
@@ -504,6 +510,32 @@ func (mgr *electionManager) verifyPendingFollower(watch watch.Watch) {
...
+			instance, err := mgr.placementManager.Instance()
+			if err != nil {
+				mgr.metrics.verifyInstanceErrors.Inc(1)
+				mgr.logError("error getting instance", err)
+				return err
+			}
+			if leaderInstance.ShardSetID() != instance.ShardSetID() {
+				err := fmt.Errorf("received invalid leader value: [%s] which owns shardSet %v, while this aggregator owns shardSet %v",
+					leader, leaderInstance.ShardSetID(), instance.ShardSetID())
+				mgr.logError("invalid leader value", err)
```
xichen2020: nit: "leader is in a different shardset group" might be a better error message.

PR #154, opened 2018-09-05
Aggregation for timed metrics

cw9:
This pr adds the API to support aggregation for timestamped metrics in M3aggregator. It can be configured to buffer past and future metrics within a time range. 

This API will not mutate the ID after the aggregation, unlike the standard untimed aggregation


aggregator/generic_elem.go
```diff
@@ -209,6 +212,23 @@ func (e *GenericElem) AddUnion(timestamp time.Time, mu unaggregated.MetricUnion)
 	return nil
 }
 
+// AddValue adds a metric value at a given timestamp.
+func (e *GenericElem) AddValue(timestamp time.Time, value float64) error {
+	alignedStart := timestamp.Truncate(e.sp.Resolution().Window).UnixNano()
+	lockedAgg, err := e.findOrCreate(alignedStart, createAggregationOptions{})
```
jskelcy: Is it possible that the `alignedStart` time is earlier than the earliest time that the `elem` will allow?

aggregator/election_mgr.go
```diff
@@ -557,6 +557,7 @@ func (mgr *electionManager) verifyPendingFollower(watch watch.Watch) {
 			continue
 		}
 		mgr.setGoalStateWithLock(FollowerState)
+		mgr.logger.Infof("goal state changed to follower")
```
jskelcy: Is this supposed to be here?

aggregator/elem_base.go
```diff
@@ -74,6 +74,16 @@ type createAggregationOptions struct {
 	initSourceSet bool
 }
 
+// IDMutationType configs if the id should be mutated after aggregation.
+type IDMutationType int
```
jskelcy: Could you leave a comment in the diff description on what is the situation when we would need to add a prefix to a metric?

aggregator/entry.go
```diff
@@ -222,12 +242,23 @@ func (e *Entry) AddUntimed(
 	}
 }
 
+// AddTimed adds a timed metric alongside its metadata.
+func (e *Entry) AddTimed(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	if err := e.applyValueRateLimit(1, e.metrics.timed.rateLimit); err != nil {
```
jskelcy: are there no additional considerations for timer metrics?

aggregator/entry.go
```diff
@@ -580,6 +613,136 @@ func (e *Entry) addUntimedWithLock(timestamp time.Time, mu unaggregated.MetricUn
...
 }
 
+func (e *Entry) addTimed(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	timeLock := e.opts.TimeLock()
+	timeLock.RLock()
+
+	// NB(xichen): it is important that we determine the current time
```
jskelcy: s/NB(xichen)/NB(chaowang)

aggregator/aggregator.go
```diff
@@ -183,6 +186,24 @@ func (agg *aggregator) AddUntimed(
 	return nil
 }
 
+func (agg *aggregator) AddTimed(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	callStart := agg.nowFn()
+	shard, err := agg.shardFor(metric.ID)
```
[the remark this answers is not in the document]
cw9: Not really, it's checked later in the path, like the `AddForward` path

aggregator/election_mgr.go
cw9: yep

aggregator/elem_base.go
cw9: Yeah we add a prefix and suffix when its a untimed aggregation, let me know if the description is not clear

aggregator/entry.go
cw9: Nope

aggregator/generic_elem.go
cw9: nope, that would be filtered out earlier

aggregator/map.go
```diff
@@ -57,6 +57,7 @@ const (
 	// nolint: megacheck
 	unknownMetricCategory metricCategory = iota
 	untimedMetric
+	timedMetric
```
jskelcy: ULTRA-NIT:
Any reason to change the enum here? I dont think there is any serialization of this value so its not a big deal, but often times things are added to the end of an enum.

cw9: Yeah there is no serialization involved so the order does not really matter, just that logically timed metric is closer to untimed metric than forward metric, so I'd rather put it next to untimed metric

aggregator/entry.go
```diff
@@ -580,6 +613,136 @@ func (e *Entry) addUntimedWithLock(timestamp time.Time, mu unaggregated.MetricUn
...
+		e.Unlock()
+		timeLock.RUnlock()
+		return err
+	}
+
+	// Update metatadata if not exists, and add metric.
+	if err := e.updateTimedMetadataWithLock(metric, metadata); err != nil {
+		e.Unlock()
+		timeLock.RUnlock()
+		return err
```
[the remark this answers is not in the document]
cw9: Not really, we would get here only if the we can't find the aggregation from `e.aggregations.index()`, so there is nothing to remove

aggregator/entry.go
```diff
@@ -580,6 +613,136 @@ func (e *Entry) addUntimedWithLock(timestamp time.Time, mu unaggregated.MetricUn
...
+	// Update the timed metadata.
+	key := aggregationKey{
+		aggregationID:  metadata.AggregationID,
+		storagePolicy:  metadata.StoragePolicy,
+		idMutationType: IDMutationDisabled,
+	}
+	listID := timedMetricListID{
+		resolution: metadata.StoragePolicy.Resolution().Window,
+	}.toMetricListID()
+	newAggregations, err := e.addNewAggregationKey(metric.Type, elemID, key, listID, e.aggregations)
```
jskelcy: NIT: not part of this diff but it seems like this function should be `addNewAggregatedKeyWithLock` to conform to norms.

cw9: Sounds good

aggregator/elem_base.go
```diff
@@ -74,6 +74,16 @@ type createAggregationOptions struct {
...
+
+const (
+	// IDMutationDisabled disables id mutation after aggregation.
+	IDMutationDisabled IDMutationType = iota
```
xichen2020: Why not have `IDMutationEnabled` as the default to maintain backward compatibility?

aggregator/follower_flush_mgr.go
```diff
@@ -338,6 +343,43 @@ func (mgr *followerFlushManager) standardFlushersFromKVUpdateWithLock(
 	return flushersWithTime
 }
 
+func (mgr *followerFlushManager) timedFlushersFromKVUpdateWithLock(
```
xichen2020: This looks pretty much the same as above, care to refactor?

aggregator/generic_elem.go
```diff
@@ -470,7 +489,12 @@ func (e *GenericElem) processValueWithAggregationLock(
 			continue
 		}
 		if !e.parsedPipeline.HasRollup {
-			flushLocalFn(fullPrefix, e.id, e.TypeStringFor(e.aggTypesOpts, aggType), timeNanos, value, e.sp)
+			switch e.idMutationType {
+			case IDMutationDisabled:
```
xichen2020: Seems more clear to call the enum `IDPrefixSuffixType` and have the values be `NoPrefixNoSuffix` and `WithPrefixWithSuffix`.

aggregator/leader_flush_mgr.go
```diff
@@ -293,6 +298,24 @@ func (mgr *leaderFlushManager) updateStandardFlushTimesWithLock(
 	mgr.metrics.standard.updateFlushTimes.Inc(int64(len(flushers)))
 }
 
+func (mgr *leaderFlushManager) updateTimedFlushTimesWithLock(
```
xichen2020: Same, refactor possibly?

aggregator/list.go
```diff
@@ -688,11 +688,87 @@ func (l *forwardedMetricList) Close() {
...
+	return fl, nil
+}
+
+func (l *timedMetricList) ID() metricListID {
+	return timedMetricListID{
+		resolution: l.resolution,
+	}.toMetricListID()
+}
+
+func (l *timedMetricList) FlushOffset() time.Duration {
```
xichen2020: Do the timed metrics need to be flushed at a fixed offset like the forwarded ones?

aggregator/options.go
```diff
@@ -767,3 +811,11 @@ func defaultMaxAllowedForwardingDelayFn(
 ) time.Duration {
 	return resolution * time.Duration(numForwardedTimes)
 }
+
+func defaultBufferForPastTimedMetricFn(resolution time.Duration) time.Duration {
+	return resolution + defaultTimedMetricBuffer
```
xichen2020: Is the default going to be 10min?

glide.yaml
```diff
@@ -118,7 +118,7 @@ import:
   version: 3ac0863d7acf3bc44daf49afef8919af12f704ef
 
 - package: github.com/gogo/protobuf
-  version: 0.4
+  version: ^1.1.1
```
xichen2020: This is backward compatible I assume?

aggregator/aggregator.go
```diff
@@ -696,6 +684,40 @@ func (m *aggregatorAddUntimedMetrics) ReportError(err error) {
...
+			"reason": "too-far-in-the-future",
+		}).Counter("errors"),
+		tooFarInThePast: scope.Tagged(map[string]string{
+			"reason": "too-far-in-the-past",
+		}).Counter("errors"),
+	}
+}
+
+func (m *aggregatorAddTimedMetrics) ReportError(err error) {
+	if err == errTooFarInTheFuture {
```
xichen2020: nit: use a switch is a bit cleaner.

aggregator/follower_flush_mgr.go
```diff
@@ -307,75 +319,41 @@ func (mgr *followerFlushManager) flushersFromKVUpdateWithLock(buckets []*flushBu
...
-		flushersWithTime = make([]flusherWithTime, 0, defaultInitialFlushCapacity)
-	)
-	for _, flusher := range flushers {
-		shard := flusher.Shard()
-		shardFlushTimes, exists := mgr.received.ByShard[shard]
-		if !exists {
-			mgr.metrics.timed.shardNotFound.Inc(1)
-			mgr.logger.WithFields(
+			metrics.shardNotFound.Inc(1)
+			logger.WithFields(
```
xichen2020: Include flush type here to distinguish standard metrics from timed metrics?

aggregator/options.go
```diff
@@ -760,17 +757,17 @@ func (o *options) initPools() {
 
 	o.counterElemPool = NewCounterElemPool(nil)
 	o.counterElemPool.Init(func() *CounterElem {
-		return MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, IDMutationDisabled, o)
+		return MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, NoPrefixNoSuffix, o)
```
xichen2020: Should the default be `WithPrefixWithSuffix`?

integration/setup.go
```diff
@@ -183,19 +183,19 @@ func newTestServerSetup(t *testing.T, opts testServerOptions) *testServerSetup {
 	counterElemPool := aggregator.NewCounterElemPool(nil)
 	aggregatorOpts = aggregatorOpts.SetCounterElemPool(counterElemPool)
 	counterElemPool.Init(func() *aggregator.CounterElem {
-		return aggregator.MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, aggregator.IDMutationDisabled, aggregatorOpts)
+		return aggregator.MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, aggregator.NoPrefixNoSuffix, aggregatorOpts)
```
xichen2020: Should the default be WithPrefixWithSuffix?

services/m3aggregator/config/aggregator.go
```diff
@@ -323,7 +323,7 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	counterElemPool := aggregator.NewCounterElemPool(counterElemPoolOpts)
 	opts = opts.SetCounterElemPool(counterElemPool)
 	counterElemPool.Init(func() *aggregator.CounterElem {
-		return aggregator.MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, aggregator.IDMutationDisabled, opts)
+		return aggregator.MustNewCounterElem(nil, policy.EmptyStoragePolicy, aggregation.DefaultTypes, applied.DefaultPipeline, 0, aggregator.NoPrefixNoSuffix, opts)
```
xichen2020: Should the default be WithPrefixWithSuffix?

aggregator/flush_times_mgr.go
```diff
@@ -366,3 +373,12 @@ func (sc flushTimesChecker) HasFlushed(
 	sc.metrics.allFlushed.Inc(1)
 	return true
 }
+
+func standardFullyFlushed(flushtimes map[int64]int64, targetNanos int64) bool {
```
xichen2020: Should this just be called `fullyFlushed` since it's used for timed metrics as well?

cw9: sure

aggregator/aggregator.go
cw9: sg

aggregator/follower_flush_mgr.go
cw9: The logger passed in already has that tag

aggregator/options.go
cw9: The usage will override it properly, but can change to `With...` instead

